PR #14, opened 2017-10-18
Setup travis

Girbons:


schollz:
@Girbons looks great! You are fast

PR #24, opened 2017-10-20
Add rate option so sender can restrict outgoing speed

### LINKED ISSUE #7: Flag to limit bandwidth for sender
This is very simple. 

1. Add a flag
2. Add a time.Sleep() after each send, which calculates based on the bandwidth and the BUFFER_SIZE

Ask me if you have questions!

heymatthew:
Closes #7

Using -rate flag, you can set the rate which croc send's files out over the wire. If not set, defaults to 1GB/s.


schollz:
Awesome thanks!

Issue #21, opened 2017-10-19
Default behavior: disconnect unless sender is available


schollz:
But this behavior could be changed to sitting until someone connects (what it is now) with `-wait`

lummie:
I'll have a look

schollz:
Awesome. Basically the idea is that if you type the wrong code phrase, and you expect someone to send you a file then it will disconnect (after a time delay to make it harder for attackers to just spam the relay server).

lummie:
oh so server side, rather than a client side -wait flag?

lummie:
So how about we adapt the potentialReceivers map to store the connection and time it connected.
Then every 30 seconds iterate this map and throw away anything that's been there longer than 1 minute?

Or other time interval that you'd like...

Lunsford94:
I had meant to take this up, but wanted to actually have everything before claiming it.

I've just made a pull request with the changes, but they're lacking any 'time delay'; I can add that as well, assuming they're good.

lummie:
@Lunsford94 go ahead and finish of :+1:

schollz:
Yeah @Lunsford94 yours looks perfect. I would just add a time.Sleep. I made a note in the code review.

PR #27, opened 2017-10-20
Adding -wait funtionality for issue #21

Lunsford94:
Changes to allow clientside '-wait' flag to wait until sender is present.

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -27,6 +27,7 @@ type Connection struct {
 	IsSender            bool
 	Debug               bool
 	DontEncrypt         bool
+        Wait                bool
 	bars                []*uiprogress.Bar
 	rate                int
 }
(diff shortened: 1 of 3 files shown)


connect.go
```diff
@@ -201,6 +208,10 @@ func (c *Connection) runClient() error {
 				message = receiveMessage(connection)
 				m := strings.Split(message, "-")
 				encryptedData, salt, iv, sendersAddress := m[0], m[1], m[2], m[3]
+                                if sendersAddress == "0.0.0.0" {
+                                        notPresent = true
```
schollz: add a time.Sleep(1 * time.Second) here to help thwart attackers who want to spam for connections

Lunsford94: Added.

schollz:
Also it looks like your entire files are in conflict? I think that you just need to run `go fmt *.go` and that should fix it.

Lunsford94:
I noticed. Older version of 'go fmt' apparently; but reformatted with newer install.

schollz:
Looks much better :) I will resolve the conflicts and merge

PR #30, opened 2017-10-21
Break from for loop if time exceeds an hour

### LINKED ISSUE #10: Break from for loop if time exceeds an hour
Specifically these two for loops: [1](https://github.com/schollz/croc/blob/91f31993ecde9a5c1c56dd92257606d4d2ce9692/relay.go#L117-L126) and [2](https://github.com/schollz/croc/blob/91f31993ecde9a5c1c56dd92257606d4d2ce9692/relay.go#L117-L126) should keep track of time and return if it exceeds some time (maybe an hour).

marcossegovia:
This should fix #10 
Any suggestion is appreciated 👍

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -166,6 +166,7 @@ func (c *Connection) runClient() error {
 	if !c.Debug {
 		c.bars = make([]*uiprogress.Bar, c.NumberOfConnections)
 	}
+	gotTimeout := false
 	gotOK := false
 	gotResponse := false
 	gotConnectionInUse := false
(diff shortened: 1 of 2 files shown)


marcossegovia:
If you want to test it @schollz, just:
- Modify the `TIMEOUT` const to a Second or so
- Open a relay
- Send a file
You'll just get the timeout as the second pass

schollz:
@marcossegovia Great! Looks perfect.

Issue #33, opened 2017-10-21
Allow sending of directories


schollz:
When specified `-send FILE` its possible that `FILE` is a directory, in which *croc* should zip the directory and transfer it as a zipped file. This is pretty easy to implement.

First it should check if `flags.File` is actually a directory, after this line: https://github.com/schollz/croc/blob/17d7b76f08e5e7025dd5fde92b5b6a3ef8ca7f82/connect.go#L51.

If it is a directory, then it should zipped it into a local temporary `.zip` file. Then change `flags.File` to the new `.zip` file and set a new `FileMetaData` variable `IsDir` (which will need to be created [here](https://github.com/schollz/croc/blob/master/connect.go#L35)) to true.

For zipping, I recommend [using mholt/archiver](https://github.com/mholt/archiver#library-use).

After this, then we can implement receiving the directory #34.

jesuiscamille:
I'll take it =)

schollz:
You got it! Let me know if you have any questions at all.

schollz:
https://github.com/schollz/croc/pull/35

PR #35, opened 2017-10-21
Automatically compress directory to a tar.gz archive

jesuiscamille:
For #33

2/4 pull request =)

### DIFF
--- a/connect.go
+++ b/connect.go
@@ -4,6 +4,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"github.com/verybluebot/tarinator-go"
 	"io"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


jesuiscamille:
Uh... Not working after 2snd try. Investigating.

schollz:
@jesuiscamille I messed up the relay about 15 minutes ago, if that was the problem...its fixed now, sorry :)

jesuiscamille:
Obviously it was me the problem...
Re-opening since i've solved the bug =)

connect.go
```diff
@@ -33,12 +34,17 @@ type Connection struct {
...
 type FileMetaData struct {
-	Name string
-	Size int
-	Hash string
-	Path string
+	Name  string
+	Size  int
+	Hash  string
+	Path  string
+	isDir bool
```
schollz: You should export `isDir` (make it `IsDir`) so it will be Marshalled into the JSON and automatically sent with the Meta data to the receiver. Otherwise looks good :)

jesuiscamille: Thanks for your review =)

PR #37, opened 2017-10-22
Adding -save and -ask-save flags, joining paths when necessary

threefjord:
Unfortunately, couldn't get the `-save` flag to function as both string and bool flag, so added `-ask-save` flag to turn on prompt for path during receive.


schollz:
Great, looks good!

PR #51, opened 2017-10-25
Added utils tests and FileSize change

afotescu:
Hey, I've added some basic tests for utils, I'm pretty sure that those are not enough cases, will try to add them later :)


schollz:
Thanks! These are great!

PR #52, opened 2017-10-25
update readme - add build, GoReport and GoDoc badges

anatolyyyyyy:
I think it's good idea - add some useful badges


schollz:
Sounds good to me! Thanks!

PR #59, opened 2018-02-07
fix file split bugs and sending server:port display error

smileboywtu:


schollz:
Thanks this looks great!

PR #60, opened 2018-02-09
add yaml config in homedir 

smileboywtu:
add yaml format config file in user home dir `.croc`


smileboywtu:
looks like ci is failed with `go get`

schollz:
Yeah, Its failing because the new `common` packages are not vendored (because its part of the *croc* package...) so not a problem. I'll fix that when its merged.

There is a problem that I don't understand. When I try to build I get this error:

```
./main.go:56:12: cannot use append(cliFlags, "github.com/codegangsta/cli".StringFlag literal) (type []"github.com/smileboywtu/croc/vendor/github.com/codegangsta/cli".Flag) as type []"github.com/codegangsta/cli".Flag in assignment
./main.go:76:20: cannot use c (type *"github.com/codegangsta/cli".Context) as type *"github.com/smileboywtu/croc/vendor/github.com/codegangsta/cli".Context in argument to common.ApplyFlags
```

I'm pretty sure codegangsta/cli has been deprecated, and there is a new version: https://github.com/urfave/cli. I've used the old library before and they had a lot of breaking changes between versions.

smileboywtu:
ok i will try to fix denpendency, then you prepare the ci test

PR #61, opened 2018-02-09
use newer version cli

smileboywtu:
use newer version cli


smileboywtu:
no reason why gox can pass the build process ?

smileboywtu:
maybe you can reference `gotty` project, i don't know why can't import package from own repo, maybe it's vendor's bugs

schollz:
This is great, its working well! Thanks!!!

PR #103, opened 2018-10-23
Improve compression speed

nicolashardy:
Use Huffman compression in order to fasten the compression speed as described in https://golang.org/pkg/compress/flate/#NewWriter


schollz:
Thanks! I forgot about Huffman

PR #115, opened 2018-11-20
Add pure Go Windows GUI

gonutz:
Hello Zack, thanks for creating this nice tool!

I have been working on a pure Go, native Windows GUI library called wui. It uses the Win32 API, can be cross-compiled easily and has no dependencies on any external libraries or DLLs, it only talks to the Windows system DLLs.

I ported your Qt Windows GUI to wui and created a Windows batch file to build it. I am not sure of your build and deployment steps, I see that you are using docker and make for it. This is why I created a file main_windows.go and a separate build.bat script. I am working on Windows and am able to run the build script and get a (6.7 MB) croc.exe file but I am not sure how this will fit into your toolchain but I hope you find this useful and can work with it.

The advantage of this is an executable that is smaller and runs both on 32 bit and 64 bit Windows from Vista up.


schollz:
@gonutz This is so cool! You are amazing, thanks!

gonutz:
Wow, that was easy :-)
Thanks to you too, for being so easy to contribute to. I appreciate your enthusiasm :-)

PR #129, opened 2019-05-10
Replace deprecated 'archive' key with new 'archives' key

meyermarcel:
With this command locally executed
```
goreleaser --rm-dist --skip-publish --skip-validate
```
no deprecation note appears.


schollz:
Again, thanks @meyermarcel !

PR #142, opened 2019-07-17
Update README.md

rain-1:
remove curl | bash recommendation

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ Status"></a>
 <a href="https://saythanks.io/to/schollz"><img src="https://img.shields.io/badge/Say%20Thanks-!-brightgreen.svg?style=flat-square" alt="Say thanks"></a>
 </p>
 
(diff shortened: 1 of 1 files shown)


schollz:
Thanks.

Is piping to bash from a https website any worse than installing a binary via `brew`/`scoop`? If so, do propose having installing from source to be the first option?

Fastidious:
I don't see the problem with the pipe to bash, but there is a [releases](https://github.com/schollz/croc/releases) page with most common platforms. Use that instead?

schollz:
What is the difference between downloading a release and piping to bash?

Fastidious:
@schollz some will argue that pipe to bash might do more that just installing a binary, without running it, specially if run as root. As I said, I don't see a problem.

jungle-boogie:
I support this PR in that it will silence the folks who have issues with piping to bash.

Issue #149, opened 2019-08-01
Install Script FreeBSD Support and Checksum Checks


TheQueasle:
I was attempting to install this on my FreeNAS box (FreeBSD) and encountered an issue when the install script gets to the checksum part.

The command `sha256sum` does not appear to be available on FreeBSD.  At least I could not find it.

While I was working on the script to fix this issue, I also noticed that it seems like the logic that does the checksum verification may not be working (lines 145-158).

```
  echo "Verifying checksum..."
  if [[ $unameu == *DARWIN* ]]; then
    checksum="$(shasum -a 256 ${dl}) $croc_file"
  else
    checksum="$(sha256sum ${dl}) $croc_file"
  fi
  checksum_check="$(cat ${dl_checksum} | grep $croc_file) $croc_file"

  if [[ "$s1" != "$s2" ]]; then
    echo "${checksum}"
    echo "${checksum_check}"
    echo "checksums are not valid, exiting"
    return 7
  fi
```

The variables `$s1` and `$s2` don't seem to be assigned anywhere.  So when the script is ran, the check `f [[ "$s1" != "$s2" ]]` will always be true.

Example run with debug output on a linux host:
```
Verifying checksum...
+ [[ LINUX == *DARWIN* ]]
++ sha256sum croc_6.1.1_Linux-64bit.tar.gz
+ checksum='87057e1a7db3dc9c91f2fb4839ab8bebaa7ad6cc31b38e84850db72dcd4b1a4d  croc_6.1.1_Linux-64bit.tar.gz croc_6.1.1_Linux-64bit.tar.gz'
++ cat croc_6.1.1_Linux-64bit.tar.gz.checksum
++ grep croc_6.1.1_Linux-64bit.tar.gz
+ checksum_check='87057e1a7db3dc9c91f2fb4839ab8bebaa7ad6cc31b38e84850db72dcd4b1a4d  croc_6.1.1_Linux-64bit.tar.gz croc_6.1.1_Linux-64bit.tar.gz'
+ [[ '' != '' ]]
+ echo Extracting...
```

A PR will follow shortly.

schollz:
This was merged! https://github.com/schollz/croc/pull/150

PR #150, opened 2019-08-01
Install Script FreeBSD Support and Checksum Checks Fixes

TheQueasle:
Adding in some fixes for FreeBSD (tested in a FreeNAS jail 'FreeBSD 11.2-STABLE') as well as a fix for the checksum logic not correctly comparing the checksums.

Before/If this PR is accepted, I plan to do more testing on this to make sure it still works on all supported platforms.

So far I've tested on FreeBSD 11.2-STABLE and CentOS 7.6.1810

Need to find a MacOS host to test on still.

### DIFF
--- a/src/install/default.txt
+++ b/src/install/default.txt
@@ -144,7 +144,7 @@ install_croc()
 
 	echo "Verifying checksum..."
 	if [[ $unameu == *DARWIN* ]]; then
-    checksum="$(shasum -a 256 ${dl}) $croc_file)"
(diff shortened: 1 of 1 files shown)


TheQueasle:
This is for issue #149

howderek:
I have tested d5a8d07d1c9005b1690d72b42d969338fe2ccc0d on macOS 10.14.6 and macOS 10.15 Beta and confirmed it installs successfully

schollz:
@howderek Thanks! I will merge it in then. 

Thanks @TheQueasle !

PR #152, opened 2019-08-13
readme: add nix install instructions

HugoReeves:
Added instructions for installing croc via the nix package manager.
Nix allows installing packages on both Linux and MacOS.


schollz:
Thanks!

PR #157, opened 2019-10-09
Adding in a more detailed error message for when an extraction tool is not found.

TheQueasle:
Ran into this when I was installing croc on a CentOS 8 minimal install.

`tar` apparently did not make the 'minimal' cut, and the old error message of

> Failed to find extraction tool in path

was too vague for my liking. :smile:

Once I installed `tar` the installer worked fine :tada:


schollz:
Thanks, I appreciate this!

Issue #172, opened 2019-11-17
Provide a source tarball with vendored dependencies


maximbaz:
Hey, would you be able to provide one more archive as part of your release, a source tarball just like Github is building for you (source.tar.gz) but with vendored dependencies? These can be generated using `go mod tidy; go mod vendor` commands.

I want this to allow building `croc` in a clean environment without internet access (assuming source tarball is already present), which might for example help with [reproducible builds](https://reproducible-builds.org/) effort.

To generate a tarball you can use `git` directly, for example Github itself is running this command (where the last argument is the tag name):

```
git archive -o v6.2.3.tar.gz --format tar.gz --prefix=croc-6.2.3/ v6.2.3
```

Let me know if this makes sense and if I can help with something!

maximbaz:
I've just went through [the same exercise](https://github.com/maximbaz/yubikey-touch-detector/commit/1bb99dfdd6a36c7c97ec3f1c27a520aed072683a) for my own project, I wasn't correct in suggesting to use `git archive`, as it won't include `vendor/` folder that is not under version control.

I can see you are using `goreleaser`, maybe you'll know a better way of achieving this, but what I found so far is based on [this article](https://medium.com/@jondot/shipping-rust-binaries-with-goreleaser-d5aa42a46be0):

1. Add post-build hook:

```
hooks:
    post: sh prepare-sources-tarball.sh
```

2. Prepare the archive in that script, I think something like this would work:

```sh
tmp=$(mktemp -d)
mkdir $tmp/croc
cp -r * $tmp/croc
(cd $tmp/croc && go mod tidy && go mod vendor)
(cd $tmp && tar -cvzf croc-src.tar.gz croc)
mv $tmp/croc-src.tar.gz dist/
```

I'm not sure based on the medium article, maybe there must be a subfolder in `dist/`, in which case perhaps worth trying:

```sh
...
mkdir -p dist/src
mv $tmp/croc-src.tar.gz dist/src/
```

Hope that helps 🙂

schollz:
@maximbaz Thanks! I got it working, but I'm not sure how to upload the prepared tarball to Github via goreleaser. Waiting on that :)

maximbaz:
I think it worked, looking at the 6.2.4 release I can see `croc-src.tar.gz`!

[...]

PR #176, opened 2019-11-19
Add version to the dir name in the source tarball

maximbaz:
Just a final touch to #172, when you unarchive the src tarball, you want the contents to be placed in a unique directory, not potentially be mixed with the unarchived old version.

### DIFF
--- a/src/install/prepare-sources-tarball.sh
+++ b/src/install/prepare-sources-tarball.sh
@@ -2,7 +2,7 @@
 tmp=$(mktemp -d)
 VERSION=$(cat ./src/cli/cli.go | grep 'Version = "v' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
 echo $VERSION
-git clone --depth 1 https://github.com/schollz/croc $tmp/croc-${VERSION}
+git clone -b v${VERSION} --depth 1 https://github.com/schollz/croc $tmp/croc-${VERSION}
 (cd $tmp/croc-${VERSION} && go mod tidy && go mod vendor)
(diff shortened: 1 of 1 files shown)


src/install/prepare-sources-tarball.sh
```diff
@@ -2,7 +2,7 @@
 tmp=$(mktemp -d)
 VERSION=$(cat ./src/cli/cli.go | grep 'Version = "v' | sed 's/[^0-9.]*\([0-9.]*\).*/\1/')
 echo $VERSION
-git clone --depth 1 https://github.com/schollz/croc $tmp/croc
-(cd $tmp/croc && go mod tidy && go mod vendor)
-(cd $tmp && tar -cvzf croc_${VERSION}_src.tar.gz croc)
+git clone --depth 1 https://github.com/schollz/croc $tmp/croc-${VERSION}
```
maximbaz: Does this hook run _after_ you tag the repo? If so, it would be better to change this to `git clone -b v${VERSION} --depth 1 https://github.com/schollz/croc $tmp/croc-${VERSION}`, to make sure this script clones the right thing (in case you decide to push new commits seconds after releasing a new version)

schollz: Yeah it does run the hook after tagging and pushing! This is a good point. It should probably do a `git clone` and then a `git checkout` in the dir.

maximbaz: The `-b` flag for `git clone` is exactly doing the `git checkout` part, but in one command 😉  I'll update this PR in a second.

schollz:
Nice trick! Didn't know that git command

Issue #177, opened 2019-11-19
Relay password is not being read from config file


maximbaz:
1. Configure a relay and save its settings to the file:

    ```
    croc --remember --relay XXX --pass YYY send file1
    ```

1. Send the file using the saved configs:

    ```
    croc send file2
    ```

Expected: it will work

Actual: relay is being read from the config, but relay password is not, resulting in:

```
could not connect to XXX: bad response: bad password
```

This works, so the problem is definitely in the password.

```
croc --pass YYY send file3
```

The password is present in the config file, it's just not being read by `croc`.

----

And if after the above I would like to use a public relay that doesn't have a password, I imagine I'd have to do the following to reset the password:

```
croc --relay 142.93.177.120:9009 --pass '' send file4
```

schollz:
Thanks, I pushed a fix for this

PR #179, opened 2019-11-19
Remember relay password on receive too #177

maximbaz:


src/cli/cli.go
```diff
@@ -334,6 +334,9 @@ func receive(c *cli.Context) (err error) {
 		if crocOptions.SharedSecret == "" {
 			crocOptions.SharedSecret = rememberedOptions.SharedSecret
 		}
+		if !c.IsSet("pass") {
```
maximbaz: I'm only not sure what's the difference to `GlobalIsSet`, I saw you used a regular `IsSet` for the `send` command, that's why I did the same

schollz: Thanks @maximbaz. I made a mistake! Both of them should be `GlobalIsSet`. You can update your PR or I can just do it :)

maximbaz: Updated 👍

schollz:
Thanks! Really appreciate that you caught my mistakes :)

PR #180, opened 2019-11-20
Support default port on receiver side

maximbaz:
Follow-up to https://github.com/schollz/croc/commit/5faacd3328a88c264f92777ec66a55f499b70354 to support receiver side as well

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -331,12 +331,12 @@ func (c *Client) Send(options TransferOptions) (err error) {
 	}
 
 	go func() {
-		log.Debugf("establishing connection to %s", c.Options.RelayAddress)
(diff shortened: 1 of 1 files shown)


src/croc/croc.go
```diff
@@ -435,11 +435,17 @@ func (c *Client) Receive() (err error) {
 				)
 				c.ExternalIPConnected = c.Options.RelayAddress
 				usingLocal = true
+				break
```
maximbaz: I noticed that this loops over the `discoveries` but never does anything to any of them except the first element, makes sense to abort early then?

schollz:
Yep, yep, this looks awesome! Thanks!!!

Issue #182, opened 2019-11-26
install fails on macOS fresh install


ferebee:
Automated install fails on fresh macOS install (10.14.6) because the directory /usr/local/bin is not created by default during macOS installation.

```
Mac:~ root# curl https://getcroc.schollz.com | bash
[…]
install: /usr/local/bin/: No such file or directory
\e[0;31m== Install attempt returned an unexpected value of 71\e[0m
```

schollz:
I can't help because I don't have a mac

Fastidious:
@ferebee I checked the install, and `/usr/localbin/` must exist for a successful install. Simply create the directory and try again.

Fastidious:
@schollz checking if the directory exists and, if not, creating it would solve this. It does exists on Linux as default, but not on macOS.

schollz:
@Fastidious But, is `/usr/local/bin` a default PATH for macOS? If not, it might be better to use a different directory (that may already exist) which is part of the default PATH.

Fastidious:
@schollz yes, it is a default path.

schollz:
@TheQueasle what do you think of changing the [install script](https://github.com/schollz/croc/blob/master/src/install/default.txt) so that it will make the install directory, if it doesn't exist? I don't mean to assign to you. But I would welcome any thoughts (or PR if you're so inclined) that you have, if you can spare the time!

TheQueasle:
@schollz I have no strong feelings against it.

I should be able to get a PR in within the next day or so (or sooner).

Oh, and feel free to ~throw me under the bus~ assign/include me in issues with the installer.  I don't mind putting my time into it :smile:

schollz:
@TheQueasle haha okay. I'd love if you can send a PR. That script is a beautiful well-thought out piece of a code and I don't want to trip it out with my hacks.

ferebee:
Everybody is exactly right about this! Sadly my skills don’t extend beyond pointing out the issue. Thanks, all.

TheQueasle:
Hit this sooner than I thought I could :tada: 

(It gave me a reason to procrastinate on doing my budget... :rofl: )

schollz:
Thanks @TheQueasle. Glad to serve you a break from your budgeting :)

@ferebee can you try it now? You can use the PR version this way:

```
curl https://github.com/schollz/croc/blob/247a6987575ad8ed27f25a3a904b2c15b69603a2/src/install/default.txt | bash
```

ferebee:
[...]

PR #183, opened 2019-11-26
Adding a function that creates the prefix

TheQueasle:
This is in response to https://github.com/schollz/croc/issues/182

Also creates logic to only create it if it does not already exist.

My only issue with this is using the `-p` flag on `mkdir`.  Something about a script auto-creating something by default makes me uneasy, though that could be just my paranoid sysadmin side coming out.

I guess so long as the script output mentions the fact it created it (the prefix in this case) then I'm more comfortable with it.

Thoughts?

OH!  And I don't have access to a Mac so this was tested with some assumptions made. It should probably get some testing on a fresh mac install.

### DIFF
--- a/src/install/default.txt
+++ b/src/install/default.txt
@@ -340,6 +340,23 @@ extract_file() {
 }
 
 #---  FUNCTION  ----------------------------------------------------------------
+#          NAME:  create_prefix
+#   DESCRIPTION:  Creates the install prefix (and any parent directories)
+#    PARAMETERS:  $1 = prefix
+#       RETURNS:  Return code of the tool used to make the directory
+#                 0 = Created the directory
+#                 >0 = Failed to create directory  
+#-------------------------------------------------------------------------------
+create_prefix() {
+  local prefix
+  local rcode
+
+  prefix="${1}"
+  
+  mkdir -p "${prefix}"
+}
+
+#---  FUNCTION  ----------------------------------------------------------------
 #          NAME:  install_file_freebsd
 #   DESCRIPTION:  Installs a file into a location using 'install'.  If EUID not
 #                 0, then attempt to use sudo.
@@ -376,7 +393,6 @@ install_file_freebsd() {
 
   return "${rcode}"
 }
-
 
 #---  FUNCTION  ----------------------------------------------------------------
 #          NAME:  install_file_linux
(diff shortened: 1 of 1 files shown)


schollz:
Nice! My only thought is that I think `sudo` might be necessary for making `/usr/local/bin`? 

I think you are right that its better that the script keeps the user informed with what its doing.

TheQueasle:
Thanks for catching the missing `sudo`!  You are definitely right there.  I'll get that fixed.

schollz:
@TheQueasle Definitely its useful. I, and probably others, like installing things without `sudo`. Were you thinking of making a home dir like `~/??/bin` and sending croc there and then adding that dir to the PATH? Or just handling a user-specified directory (which is assumed to already be on the PATH)?

TheQueasle:
I was thinking a default path of `~/.local/bin` and then informing the user that they need to update their PATH (if it is not already part of it).

schollz:
@TheQueasle Yeah, that sounds great!

ferebee:
I’m spectacularly unqualified to comment on the path issue, but an invisible executable directory feels wrong to me – it carries a stench of malware. I had to Google it, and apparently it’s a Python thing.

https://unix.stackexchange.com/questions/370943/how-did-the-local-bin-thing-start-how-widespread-is-it

I’d like to suggest ~/bin/ as I exit stage left.

schollz:
@ferebee It gets pretty wild. Other common ones are `~/.config` and `~/.settings` and `~/.cache`. All or none might be in the PATH.  Personally, I agree and I often have a `~/bin` in my computers (especially when I use Windows) but I don't know how widespread this is.

TheQueasle:
> I’m spectacularly unqualified to comment on the path issue, but an invisible executable directory feels wrong to me – it carries a stench of malware. I had to Google it, and apparently it’s a Python thing.
> 
> https://unix.stackexchange.com/questions/370943/how-did-the-local-bin-thing-start-how-widespread-is-it
> 
> I’d like to suggest ~/bin/ as I exit stage left.

I was wondering where that path came from on my systems :smile: 

While I disagree with the malware statement @ferebee , I do understand where you are coming from.

My thoughts of `~/.local` was due to my personal feelings of not liking my Home directory cluttered up and I had that path already and assumed it to be more than just a python thing.

@schollz I'm going to open an issue to track further discussion on this so we are not piggy backing off this closed PR.

PR #185, opened 2019-12-06
Support relay password for docker

xenrox:
This PR allows users to set a relay password with CROC_PASS when hosting the relay with docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,5 +10,6 @@ EXPOSE 9010
 EXPOSE 9011
 EXPOSE 9012
 EXPOSE 9013
-COPY --from=builder /go/croc/croc /croc
-CMD ["sh", "-c", "/croc --pass $CROC_PASS relay"]
+COPY --from=builder /go/croc/croc /go/croc/croc-entrypoint.sh /
+ENTRYPOINT ["/croc-entrypoint.sh"]
+CMD ["relay"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -11,5 +11,4 @@ EXPOSE 9011
 EXPOSE 9012
 EXPOSE 9013
 COPY --from=builder /go/croc/croc /croc
-ENTRYPOINT ["/croc"]
-CMD ["relay"]
+CMD ["sh", "-c", "/croc --pass $CROC_PASS relay"]
```
fbartels: this would set the password to "relay" and omit the last parameter when the user does not specify a password on his own.

Better would be a custom entrypoint script that builds up the startup argument. 

Something like:
```
#!/bin/sh
set -e
if [ -n "$CROC_PASS " ]; then
    set -- "$@" --pass "$CROC_PASS "
fi
exec croc "$@"
```

xenrox: Thanks, didn't think about that one. Should be better now.

README.md
```diff
@@ -132,7 +132,7 @@ If it's easier you can also run a relay with Docker:
 
 
 ```
-$ docker run -d -p 9009:9009 -p 9010:9010 -p 9011:9011 -p 9012:9012 -p 9013:9013 schollz/croc
+$ docker run -d -p 9009:9009 -p 9010:9010 -p 9011:9011 -p 9012:9012 -p 9013:9013 -e CROC_PASS YOURPASSWORD schollz/croc
```
fbartels: I think this is missing an = between CROC_PASS and YOURPASSWORD

xenrox: Fixed.

schollz:
@xenrox @fbartels Thanks so much, this is great!

PR #198, opened 2020-02-18
croc: secret phrase parsing

rkuprov:
what:
- `croc alpha beta gamma` is now properly understood as
 `croc alpha-beta-gamma`.

why:
- to increase convenience.


rkuprov:
This should allow entering dash-less secret phrase

schollz:
Looks great! Thanks @rkuprov

PR #209, opened 2020-04-09
relay: show version on startup

maximbaz:
Since relay is a long-running process, I find myself in situations where I dont know which version of croc I have running.

This simple startup log will make it obvious.


schollz:
Great, thanks!

PR #210, opened 2020-04-09
Add systemd service

maximbaz:
A while ago a user contributed this systemd service to Arch package, I thought it would make sense to have it available in the repo itself, so that (a) other people can find it easier and (b) if there are any improvements, the entire community will benefit of them.


schollz:
Awesome, thanks!

PR #228, opened 2020-08-22
Improve error reporting

lpar:
While I had the code checked out I decided to do some minor style cleanups and improve the error reporting.

Two tiny bugs. This is unsafe:

```
  fstat, err := os.Stat(fname)
  if fstat.Size() != fsize || err != nil {
    return
  }
```

You should check err != nil first before using the value of fstat:

    if err != nil || fstat.Size() != fsize {

The if statement will evaluate subexpressions of the same precedence from left to right.

Bug two, the unit tests were failing on dual-stack (IPv6) systems.

On to the stylistic changes...

When reporting errors, in current versions of Go

    log.Errorf("...%s", err.Error())

can be replaced with the shorter

    log.Errorf("...%v", err)

which will do the right thing, and `x/errors.Wrap` can be replaced with stdlib `fmt.Errorf`'s new `%w` error-wrapping directive, so I did that too. ref: [Go 1.13 error handling updates](https://blog.golang.org/go1.13-errors)

I mostly didn't add error checking for file close operations, with the exception of the close after successfully receiving a file. For that file close, you really want to know if (for example) the file couldn't be closed because the OS tried to flush the write buffer and failed, meaning the file is probably corrupt.

Finally, all wrapped errors should now include the actual underlying error in their message.

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -350,7 +350,9 @@ func receive(c *cli.Context) (err error) {
 		crocOptions.SharedSecret = utils.GetInput("Enter receive code: ")
 	}
 	if c.GlobalString("out") != "" {
-		os.Chdir(c.GlobalString("out"))
+		if err = os.Chdir(c.GlobalString("out")); err != nil {
+			return err
+		}
 	}
 
 	cr, err := croc.New(crocOptions)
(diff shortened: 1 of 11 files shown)


lpar:
The two are separate, I figured you might want to apply 227 but disagree with some of the style choices in this one. Git ought to be able to merge them automatically.

schollz:
@lpar Great. this is so amazing. Thank you SO much.

PR #244, opened 2020-09-06
Add Termux installation instructions to README.md

astrophena:
I added croc package (termux/termux-packages#5779) for [Termux](https://termux.com), so let's mention this in README for others.


schollz:
Cool, thanks!

PR #249, opened 2020-09-09
Update CLI to v2

jolheiser:
As title, updates the CLI library to v2.

Most things were straightforward as per [the guide](https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md).

Likely the only thing that might catch a user off-guard is [flags before args](https://github.com/urfave/cli/blob/master/docs/migrate-v1-to-v2.md#flags-before-args).

Let me know if any documentation needs to be updated, but from what I could tell everything in the README should still work perfectly fine as-is.

`go test ./...` returned a PASS :rocket:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/spaolacci/murmur3 v1.1.0 // indirect
 	github.com/stretchr/testify v1.4.0
 	github.com/tscholl2/siec v0.0.0-20191122224205-8da93652b094
-	github.com/urfave/cli v1.22.4
+	github.com/urfave/cli/v2 v2.2.0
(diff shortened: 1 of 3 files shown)


schollz:
Great, thanks! I'm excited for this. I think after this I can move towards bash completion...

It looks like you committed the croc binary. Could you delete that and rebase to get it off the history?

jolheiser:
Crap, yeah. Do you mind if I commit a gitignore as well? I was trying to avoid committing it (and obviously failed 😅 )

schollz:
Okay this looks good! Thanks @jolheiser 

Yeah, I'll take a PR with a gitignore too :+1:

PR #251, opened 2020-09-09
Simple gitignore

jolheiser:
This is the simplest gitignore I could think of that would be beneficial.

Let me know if you want anything added, such as common IDE folders (GoLand, VSCode, etc.)


schollz:
looks good!

PR #252, opened 2020-09-09
Add environment variables to CLI

jolheiser:
This PR allows a user to set environment variables for a few of the CLI options that could be considered security measures.

Let me know if you'd like me to add/rename any, as this only covers `relay`, `relay6`, and `pass` options.

I should mention I am aware of the `--remember` flag, but I figured this might be nice to avoid passwords in a plain-text file etc.


schollz:
Ooh this is great! Didn't know about this. Thanks!

PR #259, opened 2020-09-15
Add error reporting

masterZSH:
flate.NewWriter(w io.Writer, level int) (*Writer, error)
If level is not in the range [-2, 9] then the error returned will be non-nil.


schollz:
Looks good thanks

PR #260, opened 2020-09-16
Use hex.EncodeToString to encode to hex

masterZSH:
BenchmarkSha256-8

old time/op    new time/op   
477 ns/op      358 ns/op


schollz:
Cool, thanks!

Issue #264, opened 2020-09-18
Only displays the name of the first file


mig4ng:
**Describe the bug**
When sending files without the send argument (e.g. drag and drop multiple files into croc) it only shows the first file in the confirmation dialog.

**To Reproduce**
1. Use `croc file1 file2`
2. It shows the dialog with only the file1

or

1. Drag and drop multiple files into the croc executable
2. It shows the dialog with only the file1

**Expected behavior**
Show all the files.

**Version**
v8.3.2

**Additional context**
I found the code in source, unfortunately I do not know Go, therefore I cannot open a PR. With small guidance I would probably be able to open a PR. This is a small issue, no big deal, would actually be a great issue to leave open for Hackoberfest if anyone with Go skills wanted to fix it.
https://github.com/schollz/croc/blob/db388f47a6bf6a62888602d2a00f52e805fbeb56/src/cli/cli.go#L93-L99

This is not a problem if you use `croc send` however for non-tech people which use drag and drop into croc this might be confusing.

P.S. croc is really simple to use, I was amazed.

schollz:
You can't send files without the send argument.

What is happening is that `croc` is trying to figure out what you are doing and assumes you are sending a file. Your syntax `croc X` usually means to receive a file with passcode `X`. However, if `X` is a file, croc will ask you if you mean to *send* `X`. If you want to send multiple files, I suggest just doing `croc send X Y...`

mig4ng:
```
> croc file1 file2
Did you mean to send 'file1'? (y/n) y
Sending 2 files (2.1 MB)
Code is: earth-spend-talent
On the other computer run

croc earth-spend-talent

Sending (->XXX.XXX.XXX.XXX:55352)
file1     100% |████████████████████| (2.0/2.0 MB, 375.743 kB/s) [5s:0s] ✔️
file2     100% |████████████████████| (2.8/2.8 kB, 2.724 MB/s) [0s:0s] ✔️
```
I know this is a very specific edge case, yet because you send the whole `c` (context I think) in `return send(c)` it sends all the files even if you specify multiple files without the send argument. This is just a nice to have, since with the send argument it does not display the question.

It matters (e.g.) for GUI users that can select multiple files in the finder/folder and drag and drop it to croc executable/app, in Windows it tried to run the executable with the files you dragged as arguments.

[...]

PR #266, opened 2020-09-19
Show all files when sending without send command

theStack:
This PR is a possible fix for #264:

> When sending files without the send argument (e.g. drag and drop multiple files into croc) it only shows the first file in the confirmation dialog.

Through the discussion in the issue it turned out that croc always supported sending multiple files without "send" argument, but neither the code nor the message shown to the user reflects that. With the PR changeset, it is first checked that all passed arguments are indeed files/folders (not only the first one), and If they are, all of them are displayed to the user (not only the first one).

Example in master branch:
```
$ ./croc README.md Dockerfile src/croc/croc.go
Did you mean to send 'README.md'? (y/n)
```

Example in PR branch:
```
$ ./croc README.md Dockerfile src/croc/croc.go
Did you mean to send 'README.md', 'Dockerfile', 'croc.go'? (y/n)
```

Note that I'm a golang noob and the code can be very likely written in a more elegant way. Any feedback would be greatly appreciated!
Also I'm not sure If it's worth the additional LOC for a feature that is not even within the correct usage but rather a convenience feature. But in any case it made me study golang a bit and get to know new usage details about croc.

Great tool by the way!

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -89,10 +89,23 @@ func Run() (err error) {
 	app.HideHelp = false
 	app.HideVersion = false
 	app.Action = func(c *cli.Context) error {
+		allStringsAreFiles := func(strs []string) bool {
+			for _, str := range strs {
+				if !utils.Exists(str) {
+					return false;
+				}
+			}
+			return true;
+		}
+
 		// if trying to send but forgot send, let the user know
-		if c.Args().First() != "" && utils.Exists(c.Args().First()) {
-			_, fname := filepath.Split(c.Args().First())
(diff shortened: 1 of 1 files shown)


schollz:
This looks great!  

Your Go code looks perfect to me. I like the use of the inlined function (I never use those, but its a great use!). I don't think any other solution would be fewer LOC than this, its excellent.

Thank you so much for helping make croc better :)

mig4ng:
Great job fixing this, I am myself a Go noob as well however this solution looks elegant and easy to comprehend.

Thank you!

PR #280, opened 2020-10-04
CodeReview: fix resource leak && shorter code

fooofei:
1 fix a resource leak 

2 fix a possible panic of nil

3 shorter code


schollz:
This is great, thanks!!

PR #292, opened 2020-10-21
cleanup: cleanup dependencies and remove unused code

nkhang:
Hi @schollz , I want to make some contribution to this repository.

Changes:
- Use `Send` instead of `Write` since we do not check number of bytes sent.
- Remove `return` at the end of some functions since it does not return anything.
- Remove unused code.

Please take a review! This is my first contribution, feel free to comment if there is any mistake.


schollz:
Looks great! Thanks @nkhang !!

Issue #291, opened 2020-10-17
The prompt command on the sending side, which instructs a user to receive the file, is not decent


IceCodeNew:
[x] Try to [download the latest version](https://github.com/schollz/croc/releases/latest) of croc before reporting a bug.

**Describe the bug**
The prompt command on the sending side, which instructs a user to receive the file, is not decent.

**To Reproduce**
1. `croc --pass aoisora --relay "256.256.256.256:9009" send 'file.mp4'`

**Expected behavior**
Prompts command like the following:
```
Sending 'file.mp4' (15.6 MB)
Code is: foo-bar-foobar
On the other computer run

croc --pass aoisora --relay "256.256.256.256:9009" foo-bar-foobar
```

**Version**
Check `croc -version` and report it.
`croc version v8.5.0-586fc08`

**Additional context**
Currently, the prompt command is like:
```
Sending 'file.mp4' (15.6 MB)
Code is: foo-bar-foobar
On the other computer run

croc --relay "256.256.256.256:9009" foo-bar-foobar
```

If a receiving side user issues the command which prompts on the sending side as-is, the lack of the relay password will eventually lead to the `incorrect passphrase` error.

PR #293, opened 2020-10-22
fix(cli): show passphrase when user indicate --pass

nkhang:
Hi @schollz ,

This PR is to fix the bug described in #291 , please take a look.

changes:
+ use constant for default passphrase.
+ when pass is different from passphrase, show `--pass` flag in recommended receive command.


schollz:
Thanks @nkhang  this looks great! I like the use of the string builder.

IceCodeNew:
Your work is appreciated ;-)

Issue #306, opened 2020-12-08
File Size Mismatch (GiB vs GB)


hillmarcus:
**Issue**
Noticed this issue today. When beginning the transfer of a file the user is prompted with a message that includes the file's size. That size is given as a number of bytes using groupings of 1000, e.g. 1000 bytes in 1 kB. The progress bar that is displayed during the transfer makes use of another one of @schollz project's. That project displays a file's size using groupings of 1024, e.g. 1024 bytes in 1 kB. This causes the two file sizes displayed to be different.

**Proposed Fix**
The simplest fix would likely be to simply change the appropriate line (see below) from `1000` to `1024`.

https://github.com/schollz/croc/blob/5978896936b2995fb5590732a805c3658b064136/src/utils/utils.go#L139

schollz:
Great find! Would you be interested in making a PR?

PR #307, opened 2020-12-08
Issue #306, File Size Mismatch

hillmarcus:
**Should be tested before merging**
I'm not setup to compile or run Go projects, so I'm not able to test this myself.

- Updated utils.go to use a `1024` grouping instead of `1000` for file
sizes. This brings it inline with the progress bar project being used.
- Updated utils_test.go to match the changes to utils.go


schollz:
Perfect, thank you!

PR #310, opened 2020-12-17
add new cmd option to set sender ip

vzlcn0:
Add an option if the receiver already know the sender IP and want to send directly without discovery or relay,
useful in environments were multicasting and internet not available, and users do not want to run local relay.


schollz:
Looks great, thanks!

PR #315, opened 2021-01-10
Added an option to ignore attached stdin.

PeterEve:
This is to allow croc to run as a Node.js child process, as you can't stop Node piping stdio (even if you set it not to, it still pipes to /dev/null).
These changes are very helpful to me, but no worries if they don't fit the project in general.


schollz:
Great, thanks!

PR #336, opened 2021-03-06
Added support for SOCKS5 authentication (fixes #333)

### LINKED ISSUE #333: Support authenticated SOCKS5
For obvious security reasons, I'd prefer to configure my SOCKS5 proxy to require authentication rather than allow anonymous use.  However, as far as I can see by reading the manuals, the code, and testing, `croc` only supports unauthenticated SOCKS5 use.

As far as I can tell, Go supports authentication for SOCKS5 as well, but it would require a few lines of code to add the proper parameters.

hreese:


hreese:
This should not break the previous calling convention => no problems for existing users.

schollz:
Cool! Thanks @hreese

PR #337, opened 2021-03-10
src/tcp: fix dropped errors

alrs:
`salt` and `strongKeyForEncryption` are both vulnerable to being `nil` due to unhandled errors. Is this worthy of a CVE?


schollz:
awesome, thanks!

PR #338, opened 2021-03-10
src/croc: fix dropped errors

alrs:
This is the `client` sibling to https://github.com/schollz/croc/pull/337. Same question: worthy of a CVE?


schollz:
great, thanks!

Issue #231, opened 2020-09-01
too many bytes: 1765281


piegamesde:
**Describe the bug**
I tried to send a huge folder (`Sending 6122 files (11.3 GB)`) and receiving it fails

**To Reproduce**
1. croc send MyHugeFolder
2. croc password

**Expected behavior**
The files transfer correctly

**Version**
croc version v8.1.1-e250eee

**Additional context**
I'm using the Arch Linux package. Relevant log lines:

```
[debug]	11:42:52 message.go:43: writing externalip message (encrypted)
[debug]	11:42:52 comm.go:117: too many bytes: 1765281
[debug]	11:42:52 croc.go:642: got error receiving: too many bytes: 1765281
too many bytes: 1765281
```

schollz:
@piegamesde Thanks. This is an easy fix. Right now I have a limit on the max bytes in the comm just in case someone wants to send some garbage into the relay server. The relay server interprets the header and if its garbage it will see a crazy huge number there. However in your case the number is real. I'll bump it up to 4000000 and it should be okay.

whytheheckme:
Hello, I'm running into this issue today trying to transfer a 2TB folder.

schollz:
@whytheheckme what version are you using? and can you estimate the largest file in your 2TB folder? the byte level may need to be upped again. folder size itself shouldn't matter though, only the biggest file

whytheheckme:
Wow, thanks for the fast reply! The largest file I noticed on inspection was 80GB, but there could be larger. I'm attempting to move raw video files between a field shoot location and a datacenter easily. Thanks for all you do!

schollz:
Did you get that error the first time you tried transferring? Or on a subsequent try? I'm thinking that maybe it came up because of trying to sync the blocks (which is sent in one packet...which is almost always enough....except when you transfer 80gb maybe).

piegamesde:
@schollz How about fixing this issue in a more sustainable way? "Is this number reasonably small?" is not a good sanity check to tell if a client is legit or sending random bullshit. How about removing that MAX_BYTES limit altogether and instead add some other sanity checks? (Magic bytes, checksums, etc.)

Besides, the length of the transfer is some metadata the server should not care (or even know) about.

schollz:
@piegamesde You're right. I think its time to use magic bytes.

IceCodeNew:
```
Sending 80506 files (24.7 GB)
```

```
securing channel...too many bytes: 15089399
```

PR #339, opened 2021-03-14
feat: allow providing a custom --maxbytes param for a relay server

srevinsaju:
Related to #231 

making the `MAXBYTES` customizable helps to run relay servers
which can allow an arbitary amount of max bytes. By default, the
`MaxBytes` will be set to 40000.

To set a custom MaxBytes variable value:
```bash
croc relay --maxbytes 10000000
```

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -80,6 +80,7 @@ func Run() (err error) {
 			Flags: []cli.Flag{
 				&cli.StringFlag{Name: "host", Usage: "host of the relay"},
 				&cli.StringFlag{Name: "ports", Value: "9009,9010,9011,9012,9013", Usage: "ports of the relay"},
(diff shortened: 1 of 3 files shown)


schollz:
Instead of `cli.StringFlag` you should probably use the int64 flag

srevinsaju:
You are right. My bad. :smile: Rebased the fix in https://github.com/schollz/croc/pull/339/commits/304eea34fa892fc89ec72627ac3b4b6bf1d42f9a

srevinsaju:
I have looked into using int64`, but it seems that `src/comm/comm.go` uses comparisons in simple `int` instead, I have made the changes to use `cli.IntFlag` instead of `cli.StringFlag`.

schollz:
Sorry, relay is using magic prefix now. I'd still love this PR if you want to rebase and integrate it again.

srevinsaju:
@schollz rebased and pushed, https://github.com/schollz/croc/pull/339/commits/68831dc146abe5acf94a33a202bdb244ea429f32

PR #341, opened 2021-03-20
Added CROC_CONFIG_DIR env var to support custom config dir path

MatejKafka:


schollz:
Looks great, thank you!

PR #342, opened 2021-03-23
misuse of decrypt "bad password"

alfred-zhong:
Sender not receiving the "bad password" message while using a wrong password.

Misuse of `crypt.Decrypt` the "bad password". Actually it return the error but not checked.

### DIFF
--- a/src/tcp/tcp.go
+++ b/src/tcp/tcp.go
@@ -208,7 +208,10 @@ func (s *server) clientCommunication(port string, c *comm.Comm) (room string, er
 	}
 	if strings.TrimSpace(string(passwordBytes)) != s.password {
 		err = fmt.Errorf("bad password")
-		enc, _ := crypt.Decrypt([]byte(err.Error()), strongKeyForEncryption)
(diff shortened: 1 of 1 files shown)


src/tcp/tcp.go
```diff
@@ -208,7 +208,10 @@ func (s *server) clientCommunication(port string, c *comm.Comm) (room string, er
 	}
 	if strings.TrimSpace(string(passwordBytes)) != s.password {
 		err = fmt.Errorf("bad password")
-		enc, _ := crypt.Decrypt([]byte(err.Error()), strongKeyForEncryption)
+		enc, enc_err := crypt.Encrypt([]byte(err.Error()), strongKeyForEncryption)
```
schollz, line 211: Encrypt should be Decrypt?

alfred-zhong: I think `crypt.Decrypt([]byte(fmt.Errorf("bad password").Error()), strongKeyForEncryption)` will always return the `incorrect passphrase` error. Right?
Anything I missed?

schollz:
thanks! I added this in because this one is out of date

PR #344, opened 2021-03-25
Added reference to croc android-version to README.md

k3b:
See https://f-droid.org/en/packages/com.github.howeyc.crocgui/ for details


schollz:
I need to spend some time with the app before adding this, thanks.

PR #355, opened 2021-04-04
Remove dollar icons from commands without output

stingalleman:
This will make it easier to copy & paste commands. More information: https://cirosantilli.com/markdown-style-guide/#dollar-signs-in-shell-code


schollz:
Okay, seems reasonable, thanks

PR #359, opened 2021-04-08
Enhancements to bug_report.md

mizzunet:
Commented lines need to be manually removed by the user, so no need to remove those anymore
Added "big" label for this, so easily discoverable from the Issues list.
And you may label "bug confirmed" for which is confirmed


schollz:
Coool thanks

PR #361, opened 2021-04-15
Transfer symlinks without touching their targets.

tdussa:
When transfering a symlink, `croc` tries to calculate the hash of the symlinked file (i. e., the target).  This breaks when the symlink is broken.  However, there is no reason a broken symlink should not be transferable as well.  This just requires `croc` to take a hash of the symlink itself (e. g., its target), which is what this patch does.

Example:

```
$ ln -s a b; ls -l b; croc send b
lrwxrwxrwx 1 tdussa tdussa 1 Apr 15 17:12 b -> a
2021/04/15 17:12:17 stat b: no such file or directory
```

With the suggested patch, the transfer works as expected:
```
$ ln -s a b; ls -l b; croc send b
lrwxrwxrwx 1 tdussa tdussa 1 Apr 15 17:15 b -> a
Sending 'b' (1 B)
Code is: ruby-hazard-guru
On the other computer run

croc ruby-hazard-guru
```

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -324,7 +324,7 @@ func getPaths(fnames []string) (paths []string, haveFolder bool, err error) {
 	haveFolder = false
 	paths = []string{}
 	for _, fname := range fnames {
-		stat, errStat := os.Stat(fname)
+		stat, errStat := os.Lstat(fname)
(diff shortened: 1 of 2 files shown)


schollz:
This is great, thank you!!

Issue #218, opened 2020-05-18
Test fails on s390x


misery:
**Describe the bug**
Tried to build a package for Alpine. It works without problems on my machine but it fails on s390x CI.
https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/7981

**To Reproduce**
1. Build croc for s390x
2. Run unit tests

**Expected behavior**
No failing tests. ;-)

**Version**
8.0.11

**Additional context**
https://gitlab.alpinelinux.org/misery/aports/-/jobs/118553

```
>>> croc: Unpacking /var/cache/distfiles/croc_8.0.11_src.tar.gz...
?   	github.com/schollz/croc/v8	[no test files]
?   	github.com/schollz/croc/v8/src/cli	[no test files]
ok  	github.com/schollz/croc/v8/src/comm	0.319s
ok  	github.com/schollz/croc/v8/src/compress	0.016s
ok  	github.com/schollz/croc/v8/src/croc	2.752s
ok  	github.com/schollz/croc/v8/src/crypt	0.017s
?   	github.com/schollz/croc/v8/src/install	[no test files]
ok  	github.com/schollz/croc/v8/src/message	0.429s
?   	github.com/schollz/croc/v8/src/models	[no test files]
ok  	github.com/schollz/croc/v8/src/tcp	0.471s
[172.17.0.4] <nil>
--- FAIL: TestIMOHashFile (0.14s)
    utils_test.go:68: 
        	Error Trace:	utils_test.go:68
        	Error:      	Not equal: 
        	            	expected: "c0d1e123a96a598ea801cc503d3db8c0"
        	            	actual  : "c0d1e123a419ea942bc9b06c97cfb64e"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-c0d1e123a96a598ea801cc503d3db8c0
        	            	+c0d1e123a419ea942bc9b06c97cfb64e
        	Test:       	TestIMOHashFile
--- FAIL: TestHashFile (0.00s)
    utils_test.go:164: 
        	Error Trace:	utils_test.go:164
        	Error:      	Not equal: 
        	            	expected: "18c9673a4bb8325d323e7f24fda9ae1e"
        	            	actual  : "1860fd5aeb1ac95051d1cc57840037a1"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
[...]

PR #375, opened 2021-04-21
Update imohash dep

kalafut:
This may help with big-ending hashing issues (e.g. #218). It is also
more memory efficient if full file hashing is used with imohash.


schollz:
Thanks!!! Nice work. I already got it: https://github.com/schollz/croc/commit/de2bcf114c60d553bbf7a2ebfc3050130613132d

PR #376, opened 2021-04-22
Use unprivileged user in docker

Craeckie:
See 
https://americanexpress.io/do-not-run-dockerized-applications-as-root/
and 
https://engineering.bitnami.com/articles/why-non-root-containers-are-important-for-security.html


schollz:
Great, thanks!

PR #383, opened 2021-04-26
Prefer local DNS resolver over public DNS servers

### LINKED ISSUE #301: Use device's configured DNS servers instead of hardcoded DNS
Hi, I noticed that Google's DNS servers have now been [hardcoded into croc](https://github.com/schollz/croc/commit/dbc4a08ad078baec5d7c711999c2d20751f2c014). I avoid exfiltration to third-party DNS as well as Google as far as I can and would much prefer that croc sticks to the device's configured DNS servers. Would it be possible to make Google DNS optional with a switch?

oxzi:
The changes cause the local DNS resolver to be used exclusively at
first. If this fails, the public DNS servers are queried as before.

This feature was previously requested. Merging should close #301.

Along the way, three other changes are introduced:

1. A public IPv6 DNS server was added.

2. The lookup returns the first result from a public DNS server and does
   not wait for all queries to be answered.

3. In the unlikely case that no public DNS server is able to answer, an
   error will be returned.

### DIFF
--- a/src/models/constants.go
+++ b/src/models/constants.go
@@ -2,6 +2,7 @@ package models
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


schollz:
This is a really really nice PR, thank you so much @oxzi !

Issue #388, opened 2021-04-30
Use different indicator for reception of text messages


wamserma:
## To Reproduce
For the sender:
1. `echo foo > text; croc send text`
2. `croc send -t bar`

In both cases the recipient gets `Accept 'text'` and the recipient cannot distinguish between these cases unless the transfer has been initiated.

## Expected behaviour
In the text case, the recipient should see `Accept text message.` do distinguish from receiving a file, which is always quoted.

## Version
croc version v9.1.0-35106d4

schollz:
This only occurs when you send a file named "`text`", correct?

Do you have any suggestions for a different indicator?

wamserma:
> This only occurs when you send a file named "`text`", correct?
> 
> Do you have any suggestions for a different indicator?

In the simplest form, leave out the quotes, as proposed above. While this can still be overlooked, at least it can be distinguished from a quoted file name.

schollz:
Thanks, sorry I overlooked that in your issue. Sounds like a reasonable idea!

PR #390, opened 2021-05-01
Distinguish text message from file in receipient dialog

wamserma:
Fixes https://github.com/schollz/croc/issues/388


schollz:
Thanks, looks good. Though you should have `action :=` in the first instance. I can fix that though..really appreciate your help!!!

PR #401, opened 2021-06-10
refactor: Accept enter in prompts

nikoksr:
Until now, the process of receiving a file was aborted when an empty entry (`Enter`) was made. It is common under *nix that pressing `Enter` without prior input of other characters selects the default answer. This would make the workflow with croc easier.

In most usecases the default choice will be `Y`/`Yes`. The only exception is the prompt for overwriting existing files. We default to `N`/`No` in this prompt to prevent accidental overwriting of files.

Additionally I've added the folders `.idea/` and `.vscode/` to the gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ bash_autocomplete
 dist
 bin
 croc-stdin*
+
+.idea/
+.vscode/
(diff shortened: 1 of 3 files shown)


schollz:
looks great, thanks!!!

PR #409, opened 2021-08-16
Make internal DNS resolution opt-in

### LINKED ISSUE #389: Add flag for using local DNS (--internal-dns)
Thanks for creating such a great tool! I noticed that the [changelog for the most recent release](https://github.com/schollz/croc/releases/tag/v9.1.1) mentions that croc will now ["prefer local DNS resolver over public DNS servers"](https://github.com/schollz/croc/commit/8f5c73837a30f4ea232d0b096c1df6f6d512053e). This new behaviour is a great and certainly welcome improvement but using anything other than the local resolver feels like a case of NIH; I'm interested to know what the motivations behind this are.

- Why does croc fallback to a hardcoded list of recursive DNS servers? End users have no way to configure these at the command line and for some, the risk of leaking information to Cloudflare, Google, Comodo and/or Cisco might be a dealbreaker.
- Why does croc think it knows better than the user when it comes to resolving DNS queries? A failure of the local resolver may actually be desired behaviour e.g. DNS block lists.

I guess my question here is why reinvent the DNS wheel? Personally, I feel like croc should not concern itself with how DNS queries are resolved or how the IP network is configured and simply leave these things to the user/operating system.

CHTJonas:
This PR disables croc's internal DNS by default and adds an opt-in `--internal-dns` flag to bypass the local stub resolver if necessary (e.g. poorly configured ephemeral hosts).

Fixes #389.

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -80,6 +80,7 @@ func Run() (err error) {
 		},
 	}
 	app.Flags = []cli.Flag{
+		&cli.BoolFlag{Name: "internal-dns", Usage: "use a built-in DNS stub resolver rather than the host operating system"},
 		&cli.BoolFlag{Name: "remember", Usage: "save these settings to reuse next time"},
 		&cli.BoolFlag{Name: "debug", Usage: "toggle debug mode"},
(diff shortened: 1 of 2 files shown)


CHTJonas:
Cheers! 🎉

Issue #413, opened 2021-09-01
Const-ize values which are used like consts


jalavosus:
**Is your feature request related to a problem? Please describe.**
Not a **problem** per se, more of a code organization/readability/navigation thing. 

There are various values throughout the codebase which are used like constant values, examples:

https://github.com/schollz/croc/blob/master/src/croc/croc.go#L435
https://github.com/schollz/croc/blob/master/src/croc/croc.go#L406
https://github.com/schollz/croc/blob/master/src/tcp/tcp.go#L163

As well as values initialized as `var` which could/should be set as `const`, ex.: https://github.com/schollz/croc/blob/master/src/tcp/tcp.go#L39-L40

**Describe the solution you'd like**
Depending on how the value is used, store values like those shown above as `const` on a per-package or "global" level (ie. putting them in a package called `consts` or `config`). 

**Describe alternatives you've considered**
Leaving it the way it is, Croc works great regardless. 

I can start in on a draft PR later tonight which implements what I've described above for a few packages, if the devs want.

schollz:
Looks good to me! Happy to accept pr

PR #420, opened 2021-10-01
Constants and Message Types

jolheiser:
Addresses #413 

This PR also moves message types to an enum-style setup, which just makes it a bit easier to keep consistent rather than "free-form" strings.

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -33,6 +33,11 @@ import (
 	"github.com/schollz/croc/v9/src/utils"
 )
 
+var (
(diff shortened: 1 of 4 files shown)


schollz:
wonderful! thank you so much for this :) :)

maximbaz:
FYI one of the tests in `tcp_test.go` fails because it tries to modify a now constant 🙂

jolheiser:
Very odd. I'm almost positive all the tests passed when I ran them locally. 
I'll send in a fix ASAP if no one else has gotten to it.

PR #421, opened 2021-10-01
Add wildcard support for files

### LINKED ISSUE #380: Allow filename wildcards
**Is your feature request related to a problem? Please describe.**
croc send *.jpg
doesn't work

**Describe the solution you'd like**
Allow wildcards in filenames

jolheiser:
Closes #380 

[`filepath.Glob` ](https://pkg.go.dev/path/filepath#Glob) only returns filenames, so it should be safe to add to paths.  
Since the issue only referred to `*` wildcards, I search the input string for it specifically, even though `filepath.Glob` _technically_ allows for a few other matching patterns.


schollz:
looks great, thanks!!

Issue #405, opened 2021-07-09
croc can not send file to another


gengwb:
description: I just want to test croc between mac and ubantu, i send a file on mac , when i recieve the file  on ubantu, ubantu raise "securing channel...2021/07/09 14:44:37 room not ready", I check the croc version on both mac and ubantu, it is a same version which is "croc version v9.2.0-d442755" 

croc version: v9.2.0-d442755

mawoka-myblock:
Maybe try the same with the `--debug`-flag

yair-mantis:
both windows server 2019

```cxx
PS C:\ring_cli> croc --debug 8441-origami-tictac-legal
[debug] 15:21:36 cli.go:157: debug mode on
[debug] 15:21:36 croc.go:164: options: {IsSender:false SharedSecret:8441-origami-tictac-legal Debug:true RelayAddress:33e.33.133.120:9009 RelayAddress6: RelayPorts:[] RelayPassword:pass123 Stdout:false NoPrompt:false NoMultiplexing:false DisableLocal:false OnlyLocal:false IgnoreStdin:false Ask:false SendingText:false NoCompress:false IP: Overwrite:false Curve:siec HashAlgorithm:}
connecting...[debug]    15:21:36 croc.go:523: attempt to discover peers
[debug] 15:21:36 croc.go:585: discoveries: []
[debug] 15:21:36 croc.go:586: establishing connection
[debug] 15:21:36 croc.go:602: got host '33e.33.133.120' and port '9009'
[debug] 15:21:36 croc.go:604: trying connection to 33e.33.133.120:9009
[debug] 15:21:36 comm.go:59: connected to '33e.33.133.120:9009'
[debug] 15:21:36 tcp.go:439: strong key: 6a58e8e0358fb819fc78d90a8adefb8e9f058e1ab195f36e31f7a9077b3012aa
[debug] 15:21:36 tcp.go:451: sending password
[debug] 15:21:36 tcp.go:460: waiting for first ok
[debug] 15:21:36 tcp.go:475: sending room
[debug] 15:21:36 tcp.go:484: waiting for room confirmation
[debug] 15:21:36 tcp.go:497: all set
[debug] 15:21:36 croc.go:617: receiver connection established: &{connection:0xc000220040}
[debug] 15:21:36 croc.go:618: banner: 9010,9011,9012,9013
[debug] 15:21:36 croc.go:623: sending ips?
[debug] 15:21:36 croc.go:632: ips data: 
[...]

Issue #417, opened 2021-09-16
New versions do not set defaults to existing saved settings


ProactiveServices:
## Describe the bug
When upgrading croc (for example, to from 8.x to 9.3.0), new default settings are not saved if existing `receiving.json` or `send.json` exist. This causes problems such as issue #405 where I think that the `Curve` option is interpreted as null if there's no `Curve` option set in the json files.

## To Reproduce
1. Set a v8 croc install to save a setting
2. Upgrade to a v9 croc
3. Try to send a file
4. Observe error `room not ready`, which is the result of `Curve` being interpreted as null, as it is not present in old config files.

## Expected behaviour
Whether or not `--remember` is used, croc should check for existing config files and, if they require new defaults, add them.

PR #422, opened 2021-10-02
Do not override defaults with empty values

BrandonIngalls:
## About

This change will make it so a user who used the -remember
option on an older version of the application will be able
to use the latest version of the software without deleting
their saved config.

# Related

fixes schollz/croc#405
relates to schollz/croc#417


schollz:
great!!! thanks

PR #424, opened 2021-10-03
Fix TCP constant

jolheiser:
Followup to #420 

My apologies! I remember running the tests, but I must have made these changes after that. Regardless, I introduced some failing tests. This PR fixes that by changing the offending `const` back to a `var`.

cc @maximbaz


schollz:
perfect, thanks!

Issue #429, opened 2021-10-06
Switch to throttle traffic when sending file


katakombi:
**Is your feature request related to a problem? Please describe.**
I want to distribute ISO builds to a bunch of clients but I can't utilize the entire upload bandwith

**Describe the solution you'd like**
A way to throttle upload/download speed with a flag e.g. --max-bandwidth=500k

**Describe alternatives you've considered**
A traffic shaping solution e.g. wondershaper

**Additional context**

yairm210:
Since the PR was accepted, this looks closable to me

PR #432, opened 2021-10-19
added throttle upload feature

mbattista:
I added an upload throttler which was requested by https://github.com/schollz/croc/issues/429 .

If there are any mistakes in the code or points that can be improved I will try to fix them.


mbattista:
The failing test is an result of `src/croc/croc_test.go` and `src/tcp/tcp_test.go` using port `8281`.

Running only `tcp_test.go` or changing the port to e.g. ´8291´ will result in a success.
This is the same error as the one from the master branch. Should I change the port in this PR?

Or is changing the port the wrong approach and the port should be freed after the`croc_test.go`?

schollz:
Thanks no problem, this looks great!

smitty1183:
I sure this is something really simple that I am missing but no matter how I phrase it every time I use --throttleupload it returns 'flag provided but not defined: -throttleupload"

schollz:
The flag is throttleUpload , case sensitive

smitty1183:
I am sorry I was using it that way just typed it here wrong but apparently it was the order in which I was using the flag, once I put it before send and the --code flag it worked perfectly

This is how I was trying it originally

croc send --code secret -throttleUpload 2000k C:\Backups\

Once I used it like this it started working

croc --throttleUpload 2000k send --code secret C:\Backups\

I thought the whole time I was defined the value wrong.

raimocom:
. .  fell for that trap myself right now. This is not clearly communicated via `croc --help`

Expected behaviour is `croc COMMAND [OPTIONS] [FILENAME/FOLDER]`

schollz:
@raimocom if this:

```
> croc --help 

...
GLOBAL OPTIONS:
...
   --throttleUpload value  Throttle the upload speed e.g. 500k
..
```

is not clear, please submit a PR to make it more clear.

raimocom:
https://github.com/schollz/croc/pull/731

I guess i also intuitively expected --throttleUpload to be a command option of the send command. Is there any reason why it is a global option?

PR #434, opened 2021-10-27
Fixed an error

HinWaei:
Fixed an error thrown after running "curl https://getcroc.schollz.com/ | bash", which shows "install: failed to access '/etc/bash_completion.d/croc': No such file or directory". I found that the content on getcroc.schollz.com is exactly the same to this file and this error is definitely caused by nonexistence of the directory /etc/bash_completion.d/croc.

So I hope that you can update the content on getcroc.schollz.com at the same time. Thank you!

### DIFF
--- a/src/install/default.txt
+++ b/src/install/default.txt
@@ -684,7 +684,7 @@ main() {
     print_message "== Install prefix already exists. No need to create it." "info"
   fi
 
-
+  [ ! -d "/etc/bash_completion.d/croc" ] && mkdir -p "/etc/bash_completion.d/croc"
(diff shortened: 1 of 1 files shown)


UnrelatedWorks:
this should definitely be merged since I had a couple incidence where users though there was a "fatal error" when installing the program. It never induces confidence in a product when a error message appears even before you use it.

HinWaei:
@UnrelatedWorks  Thx. Just wanna make this project better as it is one of my most favourite softwares.

schollz:
Thanks this is great!!

PR #450, opened 2021-12-26
Fixed #431

### LINKED ISSUE #431: Relative folder won't send
## Describe the bug
<-- A clear and concise description of what the bug is. -->
I have the issue that folders that are given to croc by relative paths that include parent folders (e.g.: croc send ../data/testfolder/) won't send and croc leaves me with an error message "remote directory must be relative to current".
Sending any specific document within this folder is possible (croc send ../data/testfolder/test.json) and also sending the whole folder, when the working directory the command is run in does not include to switch to a parent folder (cd .. && croc send data/testfolder/).

## To Reproduce
<-- 1. Go to '...'
4. See error -->

1. Create two folders (test1, test2) and in folder test2 another folder test3
2. cd to test 1
3. send folder test3 by croc send ../test2/test3/
4. See error: remote directory must be relative to current

## Expected behaviour
<-- A clear and concise description of what you expected to happen. -->
Sending folder test3 

## Version
<-- Check "croc -v" and report it -->
croc version v9.4.2-49f50a5

## Additional context
<-- Add any other context about the problem here. -->

hamiecod:
Fixed #431 by removing the condition that requires the directory being
send to be located inside the current working directory.

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -281,10 +281,6 @@ func (c *Client) sendCollectFiles(options TransferOptions) (err error) {
 			if err != nil {
 				return
 			}
-			if !strings.HasPrefix(folderName, curFolder) {
-				err = fmt.Errorf("remote directory must be relative to current")
-				return
-			}
 			c.FilesToTransfer[i].FolderRemote = strings.TrimPrefix(folderName, curFolder)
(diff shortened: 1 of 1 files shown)


schollz:
This won't work if you send a directory beneath you...it will transfer a folder containing absolute paths

schollz:
this is fixed by #460, closing now

Issue #468, opened 2022-04-05
SIEC should be removed


samuel-lucas6:
As said in #467, [SIEC](https://github.com/tscholl2/siec) is currently the default curve, despite it being unknown and receiving insufficient cryptanalysis. It looks as if it was chosen because you're related to the designer, but that's obviously bad justification. I'm not claiming to have found a vulnerability in SIEC as I'm not a cryptographer, but common sense tells me that using such a curve is a vulnerability. Experienced cryptographers design algorithms that are broken during competitions all the time, which is why the number one rule in cryptography is to not use custom designed cryptographic algorithms. The sensible approach is to follow what cryptographers recommend by looking at modern protocols.

I'm confident many people would agree that the default curve should be Curve25519, which is why I made that other issue. If that's not something you'll implement, then at the very least, P-256 should become the default curve because it's widely used and SIEC should be removed immediately because it's used by nobody. However, it's unfortunate that the seeds for NIST curves have not been explained and may have been maliciously manipulated.

I don't understand how this problem has not been raised on GitHub considering the popularity of this project, and I don't understand your rationale for its inclusion, it being the default, and for allowing users to adjust the curve.

schollz:
Thanks @samuel-lucas6 I'll consider your comments. 

Just to be crystal clear - you are *not* claiming that you have found a vulnerability in SIEC?

samuel-lucas6:
> Just to be crystal clear - you are not claiming that you have found a vulnerability in SIEC?

I have not looked at the code for SIEC and don't have the expertise to perform cryptanalysis, so no. What I'm saying is that you've essentially violated one of the most basic principles in cryptography and somehow got away with it for this long. The curve has not received peer review from what I can tell, and there are much more popular and well researched curves to use.

You still haven't provided any justification for your design decisions. I'm very interested to hear why you thought this was a wise decision and why the curve can be adjusted. Please do not avoid these questions.

samuel-lucas6:
> SIEC is peer-reviewed (as specified in the README).

[...]

PR #469, opened 2022-04-06
Add a warning for SIEC as the default curve

realguyman:
As of right now, SIEC hasn't received third-party cryptanalysis (from searching online, I wasn't able to find anything, unfortunately). For the benefit of the community, it should be made clear that this is the case. And it should be easy to guide users to changing the default curve, hence why I included a link to jump to the appropriate section.

Seeing as #468 has taken some negative light, I want to make it clear that I have no intent of disparaging the work of `croc`, SIEC, or any persons involved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+⚠️**WARNING:** By default `croc` uses the SIEC curve (which hasn't received third-party cryptanalysis), however, it can be [changed to use standardized curves such as P-256 and P-384](#change-encryption-curve).
+
+---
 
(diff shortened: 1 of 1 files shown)


schollz:
Thanks. Would you also like to make a PR to change the default curve?

realguyman:
I'll install Go onto my system and pick up the docs and go from there. I'll open a new PR when or if I can get the new default functioning.

schollz:
Made P-256 default.

PR #478, opened 2022-05-16
Fix relay IPv4 address resolution behind DNS64

DasSkelett:
## Problem
The IPv4 address/port of the relay is resolved to `64:ff9b::5a1:458f:9009` behind resolvers that use DNS64.
When croc tries to connect to it it will fail. It should be `[64:ff9b::5a1:458f]:9009` instead, which will properly be routed through the NAT64 gateway and translated to IPv4.

## Changes
`init()` in `constants.go` now uses the stdlib `net.JoinHostPort()`, which handles this situation correctly. It should always be used when concatenating IP addresses and ports, even if the address class is (assumed to be) known.

Same change has been done to `Client.Retrieve()`, although this wasn't related to the problem.

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -778,7 +778,7 @@ func (c *Client) Receive() (err error) {
 					continue
 				}
 
-				serverTry := fmt.Sprintf("%s:%s", ip, port)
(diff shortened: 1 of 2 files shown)


schollz:
Looks great, thanks!!

Issue #275, opened 2020-09-23
Getting permission denied after receiving file


scotrod:
**After I send the file from Windows host to Linux VM I get:**
Accept 'db.txt' (0 B)? (y/n) y

Receiving (<-[::1]:1536)
[error] 2020/09/23 14:26:48 croc.go:961: open db.txt: permission denied
open db.txt: permission denied

I can transfer files from Linux box to Windows, but the reverse seems to be a permissions problem, and I don't know where to look.

**Using:**
croc version v8.0.12-7fd8fda
Windows 10 Pro 2004
Ubuntu Server 20.04 VM

schollz:
Are you using a snap?

Can you try upgrading to the latest version?

scotrod:
Hi, I'm not much of a Linux guy: `sudo snap refresh ` is what I did.
Same result after the update.

schollz:
Where is your `db.txt` located? 

It is a known issue that snap can't access dot-files in home directory (https://github.com/schollz/croc/issues/190) and unfortunately the snap folks basically told me there is no way to fix this.

schollz:
I'd be curious if the following works:

```
> cat db.txt | croc send
```

That way the snap isn't "accessing" a file.

scotrod:
`> cat db.txt | croc send` outputs: 
```
cat: db.txt: No such file or directory
open croc-stdin-821505985: permission denied
```
I searched for db.txt file with `find . -name db.txt` but that doesn't display anything. Same with locate db.txt

schollz:
Ok, I'm confused. What is this file you are sending?

```
Accept 'db.txt' (0 B)? (y/n) y
```

That means to me that you are sending an empty file named `db.txt`? Does your file exist?

scotrod:
Hello, of course, It's just an empty .txt file. I tried sending another much bigger .txt file but the result is the same.

scotrod:
Here's a video to make things more clear: [Link](https://imgur.com/MHIJtus)

schollz:
Okay, I think its a snap issue.

Try receiving the file in the `/tmp` directory or in the `/root` directory instead. If that doesn't work, try receiving the file using the pre-built version (install with `curl https://getcroc.schollz.com | bash`) or just download directly.

scotrod:
[...]

Issue #190, opened 2020-01-12
snap installed croc can't access file in home directory - permission denied


Jakethethird:
**Describe the bug**
Croc installed on Ubuntu 19.10 via sudo snap install croc or via software centre cannot access files in home directory

**To Reproduce**

```
jake@jts-800:~$ croc send Camera1_2020_01_04.avi 
open /home/jake/Camera1_2020_01_04.avi: permission denied
```

Neither cli nor store provide option to change permissions:
http://www.linuxandubuntu.com/home/snap-application-permissions

```
jake@jts-800:~$ snap interfaces croc
error: no interfaces found
jake@jts-800:~$ snap connect croc:home
error: snap "croc" has no plug named "home"
```

![20200112_SelectionNN](https://user-images.githubusercontent.com/28954472/72221403-518ce500-3552-11ea-84d9-be9e0fe52950.png)

**Version**
croc version v6.4.8-fd20b25
Ubuntu 19.10

schollz:
Thanks. I forgot to add plugs. They are there now, hopefully.

Jakethethird:
Hello, 
Now it runs but I get the following error. Is this OK?

```
jake@jts-800:~/Temp$ croc send test.txt
Sending 'test.txt' (0 B)
Code is: nylon-logo-sandra
On the other computer run

croc nylon-logo-sandra
[error]	2020/01/15 17:58:11 tcp.go:79: Error listening on :9011: listen tcp :9011: listen: operation not permitted
panic: Error listening on :9011: listen tcp :9011: listen: operation not permitted

goroutine 9 [running]:
github.com/schollz/croc/v6/src/croc.(*Client).setupLocalRelay.func1(0xc000001680, 0x897ef9, 0x4)
	/home/zns/go/src/github.com/schollz/croc/src/croc/croc.go:248 +0x157
created by github.com/schollz/croc/v6/src/croc.(*Client).setupLocalRelay
	/home/zns/go/src/github.com/schollz/croc/src/croc/croc.go:241 +0x85
```

I turned off my laptop firewall (gufw)

```
jake@jts-800:~/Temp$ croc send test.txt
Sending 'test.txt' (0 B)
Code is: pepper-chef-magenta
On the other computer run

croc pepper-chef-magenta
[error]	2020/01/15 18:00:39 tcp.go:79: Error listening on :9013: listen tcp :9013: listen: operation not permitted
panic: Error listening on :9013: listen tcp :9013: listen: operation not permitted

[...]

PR #479, opened 2022-05-20
Package as a snap (https://snapcraft.io/)

jadolg:
Snaps are very convenient for software distribution.
It'll make it even easier to install and update `croc` on systems that already use snaps like Ubuntu.

After merging this branch, sign up at https://snapcraft.io/ and link the project and GitHub repository.
Snapcraft takes care of builds and edge releases. Only stable releases need some manual interaction.

Thank you very much for this great application :-)

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,21 @@
+name: croc
+version: git
+summary: Easily and securely send things from one computer to another
+description: Easily and securely send things from one computer to another
+confinement: classic
+base: core20
+
+parts:
+  docker-image-save:
+    plugin: go
+    source: https://github.com/schollz/croc
+    source-type: git
+    build-packages:
+      - gcc
+
+apps:
+  croc:
+    command: bin/croc
+    plugs:
+      - network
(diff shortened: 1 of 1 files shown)


schollz:
snap is problematic see https://github.com/schollz/croc/issues/275 and https://github.com/schollz/croc/issues/190

I'm not interested in fielding any questions related to snap so I would only accept this if you or someone else volunteers to field all future snap related issues (basically maintain the croc as a snap)

jadolg:
Ok, I'm in 🙂

schollz:
Okay cool! I believe that the goreleaser also might need a change to add to snap- can you check on that for this PR? I have the snap credentials still and I think I can re add it

jadolg:
I was wandering around and realized the previous confinement was `strict` with some plugs to have access to `home`. I checked the conditions to use `classic` confinement and we'll never get it. Half of the snap-related issues will be related to confinement. Now I see why you decided to go away from it. I'll close this.

schollz:
Appreciate you looking into it!

Issue #487, opened 2022-07-04
Tar / zip a directory instead of sending individual files.


stefins:
**Is your feature request related to a problem? Please describe.**
Right now `croc` send all the individual files in a directory recursively which increases the time to send the directory if it contains many files. 
eg. a `.git` directory

**Describe the solution you'd like**
The solution I'm suggesting is that to tar / zip the entire directory and send it as an individual file.

**Describe alternatives you've considered**

**Additional context**

schollz:
As far as I can determine, this feature would be incompatible with being able to resume file transfers. But if you know there is a way to do it, feel free to discuss / make a PR

stefins:
We can define a cli flag to enable this feature

stefins:
I'm writing a PR for this!

schollz:
Ok cool! I was about to so thanks for letting me know. If you have any questions feel free to ask!

stefins:
Fixed in #488

nivekuil:
Tested this, much faster and convenient but it doesn't include empty directories, which seems undesirable

stellarpower:
This is really useful for sending built code as an example. We want to make sure we retain all ownership and permissions, not follow symlinks, and it's a nice feature to have one stage that creates and extracts the tarball. I'd happily forego the ability to resume a botched transfer in cases like this.

schollz:
> Tested this, much faster and convenient but it doesn't include empty directories, which seems undesirable

That sounds like a bug, would you make an issue?

PR #488, opened 2022-07-06
Added ability to zip a directory

stefins:
Related to #487

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -94,6 +94,7 @@ func Run() (err error) {
 		&cli.BoolFlag{Name: "local", Usage: "force to use only local connections"},
 		&cli.BoolFlag{Name: "ignore-stdin", Usage: "ignore piped stdin"},
 		&cli.BoolFlag{Name: "overwrite", Usage: "do not prompt to overwrite"},
+		&cli.BoolFlag{Name: "zip", Usage: "zip folder if specified"},
 		&cli.StringFlag{Name: "curve", Value: "p256", Usage: "choose an encryption curve (" + strings.Join(pake.AvailableCurves(), ", ") + ")"},
 		&cli.StringFlag{Name: "ip", Value: "", Usage: "set sender ip if known e.g. 10.0.0.1:9009, [::1]:9009"},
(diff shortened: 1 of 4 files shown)


schollz:
This looks good! I think by default the receiver should unzip the directory if it receives a zipped directory using the flag when sending. Do you want to add this as well?

stefins:
Should I send a flag to the receiver or unzip any file that ends with `.zip` ?

schollz:
A flag to the receiver. I think you already added the "TempFile" to the FileInfo so when that gets handled by the recevier they can check that.

Unzipping any zip would not be ideal because sometimes you just want to send a zip like a normal file.

stefins:
@schollz Is this ok? Should we change `TempFile` to something else?

schollz:
This is really great!! Thanks so much. 

I think just one more thing is missing - if you send a directory relative to the current directory I don't think the default behavior should be to save the zip file in the relative directory. 

I.e. if you do `croc --zip send ../SOMEFOLDER` then it will create `../SOMEFOLDER.zip` but should probably create `SOMEFOLDER.zip` in the current directory. Also the paths when unzipped should unzip to the current directory and maintain the folder structure of the folder. 

Relative paths are really annoyingly tricky like this, thats why earlier versions of croc didn't support it. Now that it supports it I want to make sure the support continues with added features. If this is getting too complicated feel free to let me know and I can help or tag in and work on this PR too.

stefins:
Thanks I will work on this further and will let you know :)

schollz:
YES ! this is so great!! I love the "Adding..." UI, very nice!

Just one more bug I found testing - `./croc --zip send ~/folder1/folder2` will attempt to send `folder1`, not `folder2`. It seems to be a slash problem because `./croc --zip send ~/folder1/folder2/` works as expected (`folder2` is sent). I'm using Ubuntu and it might be OS specific (Windows has a whole slew of slash problems) so no problem if you can't reproduce, I can fix it on my end.

stefins:
Yeah I was also having the same issue with the ending /.

stefins:
Thanks Glad I could do this! :)

PR #489, opened 2022-07-06
Add _croc zsh completion file

vapniks:


schollz:
how does this need to be installed?

vapniks:
It should be placed in one of the completion directories of `$fpath`, e.g. `/usr/share/zsh/vendor-completions`

vapniks:
I would, but I notice that there's already some code which appears to be related to completion (lines 718-721 in default.txt). However I can't find the `zsh_autocomplete_croc` file that is referenced, and `zsh_autocomplete` doesn't make sense to me. 
Is that code incomplete? and if so shall I delete it and replace it with mine?

schollz:
Yeah I think the zsh autocomplete is not as robust as yours. The current file is here: https://github.com/schollz/croc/blob/master/src/install/zsh_autocomplete

Feel free to replace it with your solution!

vapniks:
When I get the time

vapniks:
I know how to find the completions directory automatically using zsh `("${(M@)fpath:#*completions*}")`, but the install script is written in bash which means I'd have to do a series of tests for existence of certain assumed directories, which might not include the one that the user actually uses for completions. I think it might be better to just ask the user to copy the _croc file manually. What do you think?

PR #491, opened 2022-07-11
Origin/patch 1

vapniks:
Some fixes to _croc


_croc
```diff
@@ -0,0 +1,64 @@
+#compdef croc
```
captain-hoity-toity, line 1: Have you thought about the scalability of this approach?

PR #494, opened 2022-07-17
fix #493: data race in cli.relay

### LINKED ISSUE #493: data race in `cli.relay`
In the following code, there is a data race over the `err` return value.
https://github.com/schollz/croc/blob/1851327df7e882447b747fba6a45b62e9e634fe1/src/cli/cli.go#L459-L480

The goroutine(s) started at line 472 captures by reference the `err` return value and assigns to it at line 473. The same variable is (implicitly) assigned at the `return` statement at line 479.

Found with [`revive`](https://github.com/mgechev/revive)'s rule `datarace`

chavacava:
Closes #493 by declaring a new `err` variable inside the goroutine (and prevent capturing the `err` return value)

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -470,7 +470,7 @@ func relay(c *cli.Context) (err error) {
 			continue
 		}
 		go func(portStr string) {
-			err = tcp.Run(debugString, host, portStr, determinePass(c))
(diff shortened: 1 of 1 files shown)


schollz:
Thanks!

PR #514, opened 2022-10-31
pass through gofumpt

taigrr:
All code passed through gofumpt and retested. Should maintain versioned compatibility, as no public methods, variables or constants were renamed.


taigrr:
Failure is due to go 1.18's lack of support for generics, which an updated dependency (uniseg) introduces. Attempted to update it in the PR to see if that would update the workflow, but apparently only the workflow yaml on the main branch is used (which makes sense from a security perspective).

I've reverted the workflow change. Will send a separate PR for that.

taigrr:
Depends on #515

schollz:
For some reason this is failing tests....

taigrr:
Closing to replace with a better PR on this.

Issue #151, opened 2019-08-10
Suggestion: auto-retry?


omgitsraven:
I'm trying to send a large file to a friend who has a flaky connection, and while the resume function is working, we always have to manually re-connect on both ends every time something goes wrong, which means that the transfer can't continue until we're both back at our keyboards.

Specifically, on the sending end I'm seeing:
```
file.7z  28% |█████               | (3.8 MB/s) [5m8s:3h18m53s]panic: wanted to write 32814 but wrote 0: write tcp 192.168.0.187:61046->142.---.---.---:9010: wsasend: An existing connection was forcibly closed by the remote host.

goroutine 15 [running]:
github.com/schollz/croc/v6/src/croc.(*Client).sendData(0xc00003d200, 0x0)
        /home/zns/go/src/github.com/schollz/croc/src/croc/croc.go:988 +0x7cf
created by github.com/schollz/croc/v6/src/croc.(*Client).updateState
        /home/zns/go/src/github.com/schollz/croc/src/croc/croc.go:842 +0x222
```
Does that seem like the type of error that, rather than crashing, both ends could safely recognize and auto-reattempt (like once a minute or so) until the connection is re-established? Or does the NAT traversal situation mean that we'd have to go through the relay again? (Even then, while I understand that you wouldn't want TOO much traffic to the relay coming from auto-reattempts running all day, maybe trying three times before giving up or something wouldn't be too much of an extra load?)

schollz:
@omgitsraven I like this idea.

Maybe it could be a flag? Like `--reconnect`, where if its enabled it will keep reconnecting on broken connections.

omgitsraven:
That could work, although it'd either mean suffering one failed attempt in order to find out you need it, or else including the flag all the time just in case you do... Like, what would be someone's motivation *not* to use the flag just in case?

golubaca:
Enabled by default seems like an good option for me. I throws a panic, so recover will be enough to try again. You can maybe set a default flag to retry attempts (like 3 or so).

nikoksr:
Mixing all answers together would mean enabling it by default and adding a flag like `--no-reconnect` to prevent croc from automatically reconnecting.

schollz:
> @schollz I agree with @nikoksr. I would like this feature as well. Let me know what you want for this. Should the default be to auto reconnect if the transfer fails and a flag to disable? I'm still figuring out the code base, thank you.

[...]

PR #517, opened 2022-11-15
feat: first attempt at open source and go. Trying to implement an aut…

hsm2k3:
first attempt at open source and go. Trying to implement an auto-retry for issue #151. I'm not sure where to call the retry function. Looking for some constructive criticism here, thank you.

### DIFF
--- a/src/install/default.txt
+++ b/src/install/default.txt
@@ -587,6 +587,7 @@ main() {
      "x86_64" ) croc_arch="64bit";;
       "amd64" ) croc_arch="64bit";;
     "aarch64" ) croc_arch="ARM64";;
+      "arm64" ) croc_arch="ARM64";;
      "armv7l" ) croc_arch="ARM";;
        "i686" ) croc_arch="32bit";;
             * ) croc_arch="unknown";;


hsm2k3:
@schollz when you get a chance to review, the idea I was going for is to call the send function 5 more times and pass in the same values. I'm not familiar enough with the code base yet so Idk if that will actually work. I assume all the data would be in the cli.Context pointer, but my gut tells me that's not the case. Also, I'm not clear on where to call the retry func. I'm looking forward to your reply and please go easy on me I'm still learning, thank you.

schollz:
Cool! Thanks for the PR. Nice work.

One thing though is that this doesn't seem to do as described. If I use `--retry` from a sender and then have a receiver cancel the request, the sender will quit.

hsm2k3:
I didn't know receivers can cancel the request. Is [this](https://github.com/schollz/croc/blob/7f9688c7a11fb84b5e0e981216895cd20769c246/src/croc/croc.go#L832) it?

schollz:
Yes, any receiver can press Ctrl+C and cancel the request. It can happen almost anytime in the code...

lukas-mertens:
@hsm2k3 Thanks for implementing this so far. This sounds like a really good feature, please don't give up ;) I would be really glad to have it

PR #519, opened 2022-12-02
update dockerfile to v1.19

taigrr:
I missed one place to update in my previous PR. I did a global search at the time for 1.18, but apparently the dockerfile was still at 1.17.

It's my hope that this will fix the failing builds for the 386 platform I'm seeing


schollz:
great, thanks!

PR #520, opened 2022-12-05
run through gofumpt

taigrr:
hopefully this succeeds now that the dockerfile version has been updated. This is instead of #514


taigrr:
Is this the culprit? https://github.com/golang/go/issues/52919

schollz:
I'm not sure, but overall everything looks good. Something will have to be done with the Docker, but I can figure that out later. Thanks for the PR!

PR #532, opened 2023-02-10
Chmod new file based on the source file mode

### LINKED ISSUE #530: Croc should preserve the file mode of the transferred files.
Hi,
> Croc should preserve the file mode of the transferred iles.

**Is your feature request related to a problem? Please describe.**
When I'm transferring executable files using `croc` the file is written in the destination device without execution permission.
Eg. File with `-rwxrwxr-x` is written with `-rw-r--r--` on receiver end.

**Describe the solution you'd like**
We can read the file mode from the sender and create the new file with those mode in the receiver node. 

Thanks :)

stefins:
Hi,

This PR closes #530

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -1386,6 +1386,10 @@ func (c *Client) recipientInitializeFile() (err error) {
 			log.Error(errOpen)
 			return errOpen
 		}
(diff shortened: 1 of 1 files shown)


schollz:
This works great! Thanks so much.

It does open the scenario of whether a file should be updated when its permissions don't match...but I'll wait for that to become an issue.

Issue #508, opened 2022-10-04
[Feature request]: Support .gitignore file


I-Want-ToBelieve:

schollz:
Happy to accept PR

IrishMaestro:
```bash
> $ croc send .gitignore                                                                                                                                                                    [±master ●]
Sending '.gitignore' (8 B)       
Code is: [redacted]
On the other computer run

[redacted]

Sending (->[redacted])
.gitignore 100% |████████████████████| ( 8/ 8B, 7.568 kB/s)
```

I also sent a git repository and the .gitignore file was present on the remote machine. I am assuming this issue has been resolved.

PR #570, opened 2023-05-15
.gitignore support

PThorpe92:
https://github.com/schollz/croc/issues/508
Finally had time to look this over, felt pretty stupid about that last PR.  It turns out I was way over-complicating it in some ways and completely ignored obvious things like dotfiles not coming up in a filepath.Glob("*") match on *nix systems.

Finally was able to switch back, so this has been tested thoroughly on linux. 

Draft/Edit: Handling edgecase

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/chzyer/readline v1.5.1
 	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/kalafut/imohash v1.0.2
+	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/schollz/cli/v2 v2.2.1
 	github.com/schollz/logger v1.2.0
 	github.com/schollz/mnemonicode v1.0.2-0.20190421205639-63fa713ece0d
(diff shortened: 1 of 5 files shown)


schollz:
Thanks!! Simpler is always better. Give me some time to check out the PR, feel free to ping me if I forget!

PThorpe92:
Yeah I found an input case I didn't handle anyway so I'll ping you for review when I get it figured out.

PThorpe92:
I would like your opinion about the implementation of this feature, because this ultimately comes down to preference. Do you want it to ignore ALL files in the CWD that match your .gitignore? This seems like an obvious answer, however if you have a folder in your WD that is on your .gitignore list in that directory, and you try to send just that specific folder via croc, that would mean it would essentially skip it and send nothing. Or just parse the folders themselves to see if they are .git modules, then walk the sub dir and ignore those files?
currently if you "croc send *" or send anything in the cwd it will parse the top level and send only files that don't match.

PThorpe92:
@schollz  This is all set when you have time to review.

sending source folder
```
$ ./croc send ../croc
Sending 151 files and 97 folders (26.2 MB)

$ ./croc send -all ../croc                                                       
Sending 154 files and 97 folders (37.1 MB)
```

schollz:
All looks good! I would just request one thing - to make the default behavior the same as it is currently and use the flag to send only non-gitignore files.

PThorpe92:
Do you have a preference as to what the flag should be called?

PThorpe92:
@schollz  just came across this again because I saw croc on hackernews :) so I got rid of the hideous commit history. This should be all set (finally :+1: )

schollz:
Ah great! Thanks for pinging me about it - it had left my radar (whoops!).

All looks fantastic, great work.

PR #580, opened 2023-07-06
Update README.md

photosheep:
Add openSUSE installation instructions to croc

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,12 @@ On Fedora you can install with `dnf`:
 dnf install croc
 ```
 
+On openSUSE you can install with `zypper`:
+
+```
+zypper install croc
+```
+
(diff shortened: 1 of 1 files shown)


schollz:
> critical vulnerabilities

Just a nitpick here - the related bugs are of "medium" severity, not critical. Critical would be if they compromise the inherent security of the cryptography, which none of these bugs do. These bugs are related to a leak of local ip address information (#579), eliminating bad usage of reusing code phrases (#598), and  optimizations (#599, #601, #600). 

Would be great if I can get help with these issues btw.

> won't continue to exist in openSUSE

Not to be snarky, but I support this because the costs of croc are becoming burdensome again and reducing the systems its available to will help me save some costs.

Issue #584, opened 2023-07-22
Slow speed with multiple files


T3rm1:
## Describe the bug
File transmission is extremely slow if multiple files are sent, e.g. croc is used with a folder name as parameter.
I did a test with a folder containing 16 files with a total of 2.4 MB. The whole transmission took more than 10 seconds. Max upload speed of ISP is 40 Mbit/s.

How are multiple files handled in croc? Is there a significant overhead that could cause this slowdown? Single file transmission is fast.

stefins:
Could you try with the `--zip` option introduced in #488 ?

T3rm1:
That is super fast which is expected because it transfers only a single file.

stefins:
Can you provide the steps to reproduce this bug @T3rm1 ?

PR #589, opened 2023-08-17
zip file bug on windows and linux

zx9597446:
If there are directories in a zip file compressed on the Windows platform, the "\\" in the filepath will cause the files to be extracted normally on Linux, and ls will show filenames containing single quotes.

So when compressing, I replace "\\" with "/" so that it seems there are no issues on both Windows and Linux.


zx9597446:
https://github.com/schollz/croc/issues/584#issuecomment-1654530513

schollz:
Thanks!!

PR #615, opened 2023-09-25
Update README.md | Add FAQ section to README.md

alirezaalavi87:
README.md needs a FAQ section so the issues will get less flooded with trivial questions.

It needs completion of course, with more FAQs.


alirezaalavi87:
@schollz  Hi, no activity on this PR?

Issue #552, opened 2023-03-22
New releases aren't published to hub.docker.com


Ole-NM:
## Describe the bug

The latest tag points to a 2 year old release and master is v9.5.6-b7e4a73.
The last few releases and the current release (v5.6.2) aren't published on [dockerhub](https://hub.docker.com/r/schollz/croc/tags).

## To Reproduce
1. Run `podman run docker.io/schollz/croc:master --version` (or similar command with docker cli)
2. It returns `croc version v9.5.6-b7e4a73`

Alternatively go to https://hub.docker.com/r/schollz/croc/tags and sort by `Newest`.
The most recent releases were pushed ~8 months ago.

## Expected behaviour

Push a new image to dockerhub, when a new release is published.
This has worked in the past.

## Version

Versions greater than v9.5.6

## Additional context

Maybe `secrets.DOCKERHUB_TOKEN` has expired?
It is used in [.github/workflows/deploy.yml](https://github.com/schollz/croc/blob/main/.github/workflows/deploy.yml#L43-L44)
Or [something else](https://github.com/schollz/croc/actions/runs/4415457492/jobs/7738447386) failed.

I mentioned this in #242, but it isn't related to the requested (semantic) versioning.
I would like to have recent releases available via dockerhub (or some other image registry) to be able to use [podman's auto-update](https://docs.podman.io/en/latest/markdown/podman-auto-update.1.html) feature.

sjtuross:
Why is this closed?

PR #645, opened 2024-01-25
Update and fix the docker image deplyment build

rehanone:
This is a fix for #552 which looks to be closed without any fix applied.


schollz:
thanks

PR #646, opened 2024-01-25
Fix for the failing CD build

rehanone:
Unfortunately #645 still failed to publish the latest artifacts to DockerHub. The issue does not happen in local testing which leads me to believe it is to do with the legacy architectures. The logs of the failure also indicate in that direction.

I have disabled the legacy 386 and 32-bit arm build in this fix to get the normal images published. Perhaps a more detailed analysis and fix would be required for enabling the legacy builds again. The problem seams to be around these flags `-ldflags="-s -w"`

Also It looks like the CI build is using the go version 1.20 so I bumped that to be consistent.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,8 +10,8 @@ jobs:
     name: Go unit tests
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-go@v4
+    - uses: actions/checkout@v4
+    - uses: actions/setup-go@v5
       with:
         go-version: '1.20'
     - run: go version
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,11 +24,11 @@ jobs:
     steps:
       -
         name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       -
         name: Docker meta
         id: docker_meta
-        uses: docker/metadata-action@v4
+        uses: docker/metadata-action@v5
         with:
           images: schollz/croc
           # generate Docker tags based on the following events/attributes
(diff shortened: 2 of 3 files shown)


rehanone:
@schollz , Can you please review and approve this PR?

gfieni:
Hello @rehanone,

I have the same problem when building the image for the 386 architecture locally because it requires `cgo`.
The CI build failure comes from [this error](https://github.com/schollz/croc/actions/runs/7657135054/job/20867107927#step:7:547) and only concerns the 386 architecture (other builds are canceled because of it):
```
[...]
#38 6.644 runtime/cgo
#38 6.668 # runtime/cgo
#38 6.668 cgo: C compiler "gcc" not found: exec: "gcc": executable file not found in $PATH
[...]
```

Instead of removing the arm and 386 builds, the problem can be fixed by installing the `gcc` and `musl-dev` packages in the builder layer of the Dockerfile. (`RUN apk add --no-cache git gcc musl-dev`)

rehanone:
@gfieni , Thanks for the tip! It looks like with your suggested changes, all platform images are building as expects! I have updates the PR to reflect those updates and it is ready for review.

rehanone:
@schollz , Can you please provide feedback on this PR? I am concerned about the security of the current docker images as they almost 2 years old now. Hopefully, if this PR fixes the build issue then everyone who uses docker image as a relay would have a newer image.

schollz:
sounds good!

pepa65:
What about the .deb targets? It seems they no longer get released.

schollz:
@pepa65 you can add the .deb targets you want into the new release.yml: https://github.com/schollz/croc/blob/main/.github/workflows/release.yml. I'll accept the pr

pepa65:
Hmm, no more goreleaser... It's not immediately clear to me how to add this in. I will wait for someone to pick this up. In the meantime, I am happy to just get the binaries I need, which is what I love about go and croc!

PR #671, opened 2024-02-22
Build release binaries for DragonFly/Free/Net/OpenBSD

dbohdan:
This PR modifies the release workflow to build binaries for DragonFly BSD, FreeBSD, NetBSD, and OpenBSD. The updated workflow builds binaries for AMD64 and ARM64 for those systems that support them. DragonFly BSD only supports AMD64, so only one binary is built. For NetBSD, it also builds an i386 binary because a major use of NetBSD is retrocomputing.

The workflow succeeds on my fork of the repository: https://github.com/dbohdan/croc/actions/runs/8011298184, https://github.com/dbohdan/croc/releases/tag/release-test. (The test branch includes an extra commit to change the repository URL in the workflow file to mine. The commit isn't part of this PR.)


schollz:
awesome, thanks!

PR #683, opened 2024-03-30
Support spaces in 4 word codes

a1lu:
Since the code was extended by the 4 digit block, the patch from https://github.com/schollz/croc/pull/198 did not work anymore.


a1lu:
I am not sure if the 4 digit block is always present. If so, it would be enough to replace `case 3:` by `case 4:`.

schollz:
thank you!

Issue #681, opened 2024-03-27
Croc fail to run when HOME env var is not defined


sivachandran:
## Describe the bug
<-- A clear and concise description of what the bug is. -->
Croc fail to run with the following error if `HOME` env var is not defined
```
[error] 2024/03/27 08:19:11 cli.go:448: $HOME is not defined
2024/03/27 08:19:11 $HOME is not defined
```

## To Reproduce
```
$ HOME= ./croc
[error] 2024/03/27 08:19:11 cli.go:448: $HOME is not defined
2024/03/27 08:19:11 $HOME is not defined
```
<-- 1. Go to '...'
4. See error -->

## Expected behaviour
<-- A clear and concise description of what you expected to happen. -->
Croc should run normally

## Version
<-- Check "croc -v" and report it -->
```
$ ./croc -v                                                          
croc version v9.6.14
```

## Additional context
I tried to run Croc on hardened OS and embedded device where `HOME` is not defined to prevent creation of user config files. Croc fails to run with the above error. I have to manually define the `HOME` env to run Croc, like `HOME=/tmp ./croc`. It would be great if Croc doesn't require `HOME` env var, like a portable executable that doesn't create any file on the host machine.

a1lu:
Hi,
I created the pull request https://github.com/schollz/croc/pull/684 so that croc does not crash if `$HOME` is not set, but one of `$CROC_CONFIG_DIR` or `$XDG_CONFIG_HOME`.
Does this solve your problem, or are the latter two environment variables also not set on your system?

PR #685, opened 2024-03-30
Gracefully handle non existend receive config file

a1lu:
This PR is based on https://github.com/schollz/croc/pull/684

As I understand the handling of `send.json` in the  `send` function, it prints but ignores an error of `utils.GetConfigDir()`. I tried to copy this behavior to the `receive` function:

If `utils.GetConfigDir()` fails, the error is printed and returned to `receive`. Though, croc only exits if `c.Bool("remember")` is `true`, because only then, the config file is really needed.

This may be related to #681

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -153,13 +153,22 @@ func setDebugLevel(c *cli.Context) {
 	}
 }
 
-func getConfigFile() string {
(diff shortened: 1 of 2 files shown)


schollz:
thanks, this looks great!

PR #686, opened 2024-04-01
Avoid to create config dir if not required

a1lu:
The config directory must only exist if `--remember` is set.

Otherwise croc only tries to read the config file on the returned path, that already could be `""` if 	`utils.GetConfigDir()` returns an error and if the config file does not exist during read, errors are already silently ignored. 

Edit: should fix https://github.com/schollz/croc/issues/681


schollz:
perfect, thanks!

PR #700, opened 2024-05-20
fix: establish encryption layer before transfering ip information

### LINKED ISSUE #597: Unencrypted "ips?" Message Leaks Information from the Sender Side
As a typical part of the Croc protocol (if no explicit `--ip` is passed), the receiver will ask the sender about its locally assigned IP addresses via the `ips?` message (croc.go:792). This message and its reply are sent unencrypted. I assume there is no encryption, because the receiver might still switch the connection to a direct one, without going through a public relay, and setting up the encryption context twice might add additional latency, or additional code complexity.

The message being unencrypted means, however, that the sender will send out cleartext information over the Internet, containing all locally assigned IP addresses. This might be an unexpected information leak for a range of users. It can reveal information about the structure of internal networks or otherwise offer information about the identity of the sender.

To fix this, the encryption layer should be established before any other data is transferred between sender and receiver.

schollz:
Fixes #597 

@mgerstner please review and advise

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -81,6 +81,11 @@ type Options struct {
 	GitIgnore      bool
 }
 
+type SimpleMessage struct {
+	Bytes []byte
+	Kind  string
+}
+
 // Client holds the state of the croc transfer
 type Client struct {
 	Options                         Options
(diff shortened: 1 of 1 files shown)


schollz:
Fixes https://nvd.nist.gov/vuln/detail/CVE-2023-43618

PR #749, opened 2024-07-03
Refactor tcp

Ozoniuss:
I'm still working on #743, but while trying to figure out if the change works in all cases (still not 100% convinced I fully understand what happens when the sender starts a local relay, and haven't yet considered peer discovery), I saw some potential stylistic improvements for the TCP implementation. This PR does not change any functionality, and is mostly focused around the configuration of the TCP listener. Each individual commit will describe the change in more detail.

I also added a diagram for what I think happens when connecting to the local relay. I drew it for myself initially, but thought it might help other contributors.

Let me know if you like the refactor and want me to extend it further (e.g. refactor the `Run` function completely, or tests) The way I wrote the refactor was meant to change the least amount of files possible.

Tests pass on linux. On windows they fail on the main branch too, probably because the networking is implemented differently.


Ozoniuss:
Should be ready, the only functional change is room removal (using different constructs, it's supposed to work the same). But I have some free time tomorrow, can test the new version in more network conditions than just local, just to be extra sure nothing breaks. I'll let you know once I've done that

Ozoniuss:
Given the room stopping isn't actually called, I think this change should be functionally equivalent to the previous code and is ready to be merged. I'd still keep the deletion mechanism, since I believe it would work for shutting down gracefully that goroutine, when that would be implemented. I also added some debug logs for the cleanup.

PR #753, opened 2024-07-09
Add receiving on Linux and classic mode to README

clemisch:
This adds a mention of receiving on Linux and `--classic` mode to the README.


clemisch:
It may be better longterm to add separate sections about `--classic` mode and Linux behavior. I didn't want to make too large of a change / clutter the README. Now it at leasts mentions classic mode somewhere with a link to the CVE.

PR #757, opened 2024-07-22
Avoid leaking process using pipe (echo secret |)

01-1:


01-1:
Forgot to merge latest commit

x64x2 (approved):

PR #758, opened 2024-07-22
Use echo + pipe to avoid leaking secret in process name

01-1:
Using `echo [secret] | croc` avoids leaking the secret via process name, and is more concise than using an environment variable. (Additionally, exporting an environment variable still has the risk of leaking the secret if the user runs another program )

This pull request changes the cli messages to use this.
I also fixed a typo in the cli error messages.

actually
CROC_SECRET=*** is better i am going to change this


01-1:
see other PR

PR #759, opened 2024-07-22
Change CLI messages to use CROC_SECRET

01-1:
This pull request changes the cli messages to use CROC_SECRET.
Previously, the error message used export CROC_SECRET, which is unnecessary as you can just run `CROC_SECRET=**** croc` in one line.
Also, the messages for the sender used the classic command:
```
On the other computer run

croc ****
```
This made it incompatible with UNIX unless classic mode was activated.

The new command is
```
CROC_SECRET=**** croc
```

I also fixed a typo in the cli error messages.

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -646,7 +646,7 @@ func (c *Client) Send(filesInfo []FileInfo, emptyFoldersToTransfer []FileInfo, t
 	if c.Options.RelayPassword != models.DEFAULT_PASSPHRASE {
 		flags.WriteString("--pass " + c.Options.RelayPassword + " ")
 	}
-	fmt.Fprintf(os.Stderr, "Code is: %[1]s\nOn the other computer run\n\ncroc %[2]s%[1]s\n", c.Options.SharedSecret, flags.String())
+	fmt.Fprintf(os.Stderr, "Code is: %[1]s\nOn the other computer run\n\nCROC_SECRET=%[1]s croc %[2]s\n", c.Options.SharedSecret, flags.String())
 	if c.Options.Ask {
 		machid, _ := machineid.ID()
 		fmt.Fprintf(os.Stderr, "\rYour machine ID is '%s'\n", machid)
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -646,7 +646,7 @@ func (c *Client) Send(filesInfo []FileInfo, emptyFoldersToTransfer []FileInfo, t
 	if c.Options.RelayPassword != models.DEFAULT_PASSPHRASE {
 		flags.WriteString("--pass " + c.Options.RelayPassword + " ")
 	}
-	fmt.Fprintf(os.Stderr, "Code is: %[1]s\nOn the other computer run\n\nCROC_SECRET=%[1]s croc %[2]s\n", c.Options.SharedSecret, flags.String())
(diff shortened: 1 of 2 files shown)


01-1:
Not sure if CROC_SECRET works on all platforms though (windows?)

src/croc/croc.go
```diff
@@ -646,7 +646,7 @@ func (c *Client) Send(filesInfo []FileInfo, emptyFoldersToTransfer []FileInfo, t
 	if c.Options.RelayPassword != models.DEFAULT_PASSPHRASE {
 		flags.WriteString("--pass " + c.Options.RelayPassword + " ")
 	}
-	fmt.Fprintf(os.Stderr, "Code is: %[1]s\nOn the other computer run\n\ncroc %[2]s%[1]s\n", c.Options.SharedSecret, flags.String())
+	fmt.Fprintf(os.Stderr, "Code is: %[1]s\nOn the other computer run\n\nCROC_SECRET=%[1]s croc %[2]s\n", c.Options.SharedSecret, flags.String())
```
schollz, line 649: This isn't necessary on Windows. Can you change it to only print this on Unix systems (you can use `runtime.GOOS == "windows"`)

01-1: The receiver system may not be the same as the sender. I changed this to print both commands.

schollz:
CROC_SECRET is not needed on Windows, so please keep the regular print statements for Windows systems (if `os.Runtime=="windows"`) and have your new print statement for non-windows

x64x2 (approved):

schollz:
@01-1 does the suggested change make sense?

01-1:
No, it doesn't - [as I said](https://github.com/schollz/croc/pull/759#discussion_r1688279636), the receiver system may not be the same as the sender. For example, a windows system could send a file to a linux system, and vice versa.

schollz:
I think its best to revert line 649 back to original. The receiver should interact with the messages

01-1:
It doesn't make sense to tell the user to run a command that will not work by default

schollz:
I think we can all agree that no one agrees on user interaction. 

to me, the interaction makes sense because you type in `croc PASSPHRASE` and you get a explainer of all your options:

```
On UNIX systems, to receive with croc you either need
to set a code phrase using your environmental variables:

  export CROC_SECRET="****"
  croc

Or you can specify the code phrase when you run croc without
declaring the secret on the command line:

  croc
  Enter receive code: ****

Or you can go back to the classic croc behavior by enabling classic mode:

  croc --classic
```

versus having to have that information in the sender's purview also.

01-1:
OK. It is much easier to be able to copy-paste the command though. Maybe add an option to print the unix command?

schollz:
I mean, its pretty easy just to type `croc` on your receiver, press enter, and copy paste the password? if you are on a single computer, thats really the easiset

nhtzr:
Hi. I arrived here because I wanted to propose a similar change. I believe a workflow that easily and quickly lets you do the right thing is valuable as it reduces the amount of frustation of anyone involved.

However I think the proposed message is hard to read, I would prefer something like:
```
Sending 'a.txt' (2.6 kB)         
Code is: 1106-yoyo-genius-patent

On the other computer run:
(For Windows)
    croc 1106-yoyo-genius-patent
(For Linux/OSX)
    CROC_SECRET="1106-yoyo-genius-patent" croc
```

Code would be
```
	fmt.Fprintf(os.Stderr, `Code is: %[1]s

On the other computer run:
(For Windows)
    croc %[2]s%[1]s
(For Linux/OSX)
    CROC_SECRET=%[1]q croc %[2]s
`, c.Options.SharedSecret, flags.String())
```

schollz:
@nhtzr please submit a PR!

Issue #711, opened 2024-05-24
Extend "--transfers" flag to relay in addition to send


jathek:
https://github.com/schollz/croc/pull/660 by @qk-santi added a `--transfers` flag to specify the number of ports for croc send. Croc relay still uses manual ports though. It would be great if relay could use the --transfers flag as well.

croc 10.0.2 relay -h, showing old --ports flag and no new --transfers flag:
![image](https://github.com/schollz/croc/assets/692914/f8f7cc90-6bef-4cef-8ab7-2c468f25925b)

PR #660, opened 2024-02-12
define ports by amount, not individually

qk-santi:
Defining the number of ports for transfer connections is a bit cumbersome, especially if you want to create quite a few ones
```
croc send --ports 9009,9010,9011,9012,9013,9014,9015,9016,9017 my_file.gzip
```

Switched to:
```
Define base port (if different than default) + number of parallel connections
croc send --port 9009 --transfers 8 my_file.gzip
croc send --transfers 8 my_file.gzip
```

Fixes #617

PR #763, opened 2024-07-27
define relay ports by amount

tnothy:
Switched to defining ports for relay by amount similar to croc send implemented in [#660](https://github.com/schollz/croc/pull/660)

Fixes [#711](https://github.com/schollz/croc/issues/711)


schollz:
this is fine, but backwards compatibility is critical so please alter the PR to still allow the `ports` flag

tnothy:
restored the `ports` flag

PR #766, opened 2024-07-28
fix test case name

tnothy:
Seems like the test was ignored when running `go test` due to lowercase `test`.

Results of `go test -json ./src/croc/ | jq -s 'map(select(.Test and .Action == "pass")) | length'`
Before: 6
After: 7


x64x2 (commented):
Doesn't fix anything

tnothy:
I believe that right now, with the test function starting with a lower case letter the test is not executed

PR #797, opened 2024-09-01
Contains to HasPrefix

### LINKED ISSUE #796: can't send files with `..` in filename
receiving side:
```
Enter receive code: [...]
securing channel...2024/09/01 07:42:35 basename cannot contain '..': '[...]'
```

croc version v10.0.11

asukaminato0721:
may be fix #796


schollz:
Good start, it must check whether ".." near any path separator (which are different on windows and unix)

PR #813, opened 2024-09-19
Adding shell completion script for zsh

Eloquencere:
Added shell completion script for zsh shell.

### DIFF
--- a/completion/_croc.zsh
+++ b/completion/_croc.zsh
@@ -1,0 +1,121 @@
+#compdef croc
+
+_croc() {
+	local line state
+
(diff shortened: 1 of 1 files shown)


Eloquencere:
hi @schollz this PR has remained inactive for a while, could you confirm if this is still under consideration to get merged?

x64x2:
this has already done by someone else

Eloquencere:
Okay, that's good but, when I install croc I don't have any shell completion and I can't seem to find it in the file structure of the repo, it's possible that I missed it. So, what's the issue?

PR #822, opened 2024-10-06
Added copy to clipboard functionality for windows,linux and mac

elliot40404:
Added copy to clipboard functionality for Windows, Linux and mac using `os/exec` package. Please LMK if I need to change anything else. Also my formatter moved around 1 or 2 lines, I hope that won't be a problem. Thanks

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -12,15 +12,14 @@ import (
 	"math"
 	"net"
 	"os"
+	"os/exec"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	"golang.org/x/term"
-	"golang.org/x/time/rate"
 
 	"github.com/denisbrodbeck/machineid"
 	ignore "github.com/sabhiram/go-gitignore"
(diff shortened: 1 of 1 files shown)


ryder203:
Very nice and it's cool that you implemented it for MacOS and Linux too, but is it problematic if the used Linux distro does not have xclip installed? Would there be the need of some error output if xclip is not present? I'm not into Linux that much and maybe it's a stupid question.

elliot40404:
@ryder203 Your question is valid, xclip or for that matter any of the other utils used maybe not be present  or deleted or just not accessible for any reason, in that case it will fail silently as the copy to clipboard functionality is not essential for croc to work. If it fails the l line "code has been copied to clipboard" will not be printed to std err and if used with --debug it will also show  up in the logs.

elliot40404:
My bad. Now all tests should pass.

schollz:
thanks @elliot40404 !!! great pr.

ryder203:
Awesome. Now using croc or croc tooth has some more flow with the next version on windows. A friend and I tested croc these days and I sent him a 20GB file in like 55 minutes and all worked like a charm. I really love it.

bitraid:
I'm sorry for commenting on an already merged PR, but I have a couple of suggestions:
- On Windows: There is no need to run `clip` through `cmd`.
- On Linux: Should support Wayland, by checking for example `$XDG_SESSION_TYPE` and using `wl-copy` instead of `xclip`

PR #824, opened 2024-10-12
Improve clipboard copy

bitraid:
A couple of small modifications to the new clipboard-copy function:
- Don't use `cmd` to run `clip` on Windows (it's not an internal command).
- Add Wayland support on Linux (by using `wl-copy` after checking `$XDG_SESSION_TYPE`).


schollz:
Looks great thanks!!!

x64x2 (approved):

Issue #809, opened 2024-09-16
Issue when sending files/folders on a different drive letter


ymich9963:
For example if I try sending a folder on a PCs E: drive, and the receiver doesn't have an E: drive, it doesn't successfully send. I am on Windows 10 on both PCs.

schollz:
thanks, will accept PR

sk25469:
Hi, I would like to work on this.

schollz:
please do! thank you @sk25469

sk25469:
@schollz what kind of behavior do we want here? I think we can do something like:

1. Check if the recipient has the destination path, then send it
2. If not, then we can send it to the Download folder or the current folder

schollz:
The received should receive in the folder they ran the croc program

PR #836, opened 2024-11-10
Bug Fix: Fixed error when sending files to a target directory that didn't exist

sk25469:
- Fixes [Issue #809](https://github.com/schollz/croc/issues/809), which caused errors when attempting to send files or folders located on a different drive letter on Windows systems.


schollz:
does this fix #809? i'm pretty sure #809 can happen even if the `out` flag is not used

sk25469:
Sure thanks for letting me know.
I will look for other ways, but can you help here with what other cases it might fail in? We can send to a specific directory only with the `--out` flag, right?

schollz:
@ymich9963 can provide guidance

ymich9963:
Sure I'll try to run it on the same machines and let you know what happens. Do I just use a flag to specify the output or does it get saved in the pwd?

ymich9963:
I don't remember. I don't think I specified. I was just on the E:/ drive in one PC and on the C:/ drive in another. Could you specify the output for croc before I raised the issue? I remember that it would just output to the same directory as the source.

ymich9963:
Should I build from source to test this?

sk25469:
Yes u should build from source.
Earlier if u used `--out` and the folder didnt exist, it would just fail.
If u used without the flag, it would just send in the pwd

PR #591, opened 2023-08-30
Add printing receive command as a qr code with new --qrcode flag

lukenvn:
Hi guys, 

This pull request adds a new feature to the Croc software, allowing users to display the receive command as a QR code by using the --qrcode flag. This feature enhances the usability and convenience of Croc by providing an easily scannable QR code representation of the receive command.I also plan to add scan qrcode feature in croc android soon.

Changes Made:
1. Added a new command-line flag --qrcode to the Croc software.
2. When the --qrcode flag is used, the receive command is extracted and converted into a QR code using the go-qrcode library.
3. The QR code is displayed to the user in the command-line interface.
image

Benefits:
1. Improved user experience: Users can now conveniently display the receive command as a QR code, making it easier to scan and share.
2. Enhanced accessibility: The QR code representation allows users to use QR code scanning applications or libraries to automate the process of capturing the receive command.
3. Increased efficiency: By providing a direct QR code display option, users can quickly access the receive command without manually typing or copying it.

P/s: This is my first time of using go lang so Im not sure what i'm is in good practice, so please let me know if i should improve something. Thank you in advanced .

PR #847, opened 2024-11-24
Add a qrcode flag

Ozoniuss:
- Introduces a new flag which allows displaying the receive command using a qr code.
- Adds a newline after the message indicating that the code was copied to the clipboard.

This is an exact copy of the feature introduced in https://github.com/schollz/croc/pull/591, which went stale. The only addition is a newline.

image


schollz:
Thanks!!!

jpph:
Thanks!!!

PR #881, opened 2025-01-31
Bump Go, dependencies, CI - Upload artifacts after running tests - Update readme

blu3berryys:
Also adds a comment to stale action (there wasnt a proper comment at all.)

also you might want to change stale times

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,1 @@
-# These are supported funding model platforms
-
 github: schollz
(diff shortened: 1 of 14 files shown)


schollz:
Thanks, lots of great work here. Did you do this by hand or use an LLM? Just curious.

blu3berryys:
I didnt use a llm

blu3berryys:
Also I would recommend you to change the stale action messages and its time. Its defaults are both 60 for pr and issue

schollz:
thanks. what is the recommendation for? to be longer or shorter? why?

blu3berryys:
it can be shorter aswell. also you can change the message

sergeevabc:
Still works under Windows 7. Hooray!

PR #916, opened 2025-04-20
Added traefik 3.0 setup steps

OisinHick:
Hey there!

Really love this project and wanted to set it up to run behind my VPS using Traefik 3.

I have added steps to put croc behind traefik. This took me a bit of work to get up and running perfectly, and given traefik is a popular proxy, I wanted to share this with the community.

I'm aware there is alot of users just editing readmes and then commiting, so I can also submit the traefik.yml and croc.yml files to an examples folder upon @schollz request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,6 +285,100 @@ To send files using your custom relay:
 croc --pass YOURPASSWORD --relay "myreal.example.com:9009" send [filename]
 ```
 
+#### Self-host Relay with Docker behind Traefik 3.0
+
+You can also run a relay with Docker behind traefik 3.0. Create a croc.yml as shown:
+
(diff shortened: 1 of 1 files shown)


vaygr:
TLS part should be removed, it's unsupported configuration. Has nothing to do with cloudflare.

I have Let's Encrypt configured, and on initial send I get the error:

```shell
could not connect to ...: initial bytes are not magic: 48545450
```

vaygr:
@OisinHick yes I would remove any mentions regarding TLS. The other ports part really depend on the outcome of the discussion in the referenced issue. I'm not convinced that traffic is going through the relay with this configuration.

Issue #968, opened 2025-09-24
Is there a Health Check Path available?


felipewnp:
Hi there!

First of all, thanks for the effort to keep this amazing project!

I would like to know if there's a path I could probe for a health check on a Docker/Kubernetes environment for the self-hosted relay server.

I've tried reading the source code, trying to find any path at all to check, but I know nothing about Go and couldn't find any.

Again, thanks for the amazing project!

sansneo:
Would something like this in Dockerfile do?
```
HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
    CMD nc -z localhost 9009 || exit 1
```
Do you want an actual HTTP /health endpoint? That might not be what @schollz wants.
Reading through the source code, let us know @felipewnp!

felipewnp:
> Would something like this in Dockerfile do?
> 
> # Simple TCP health check with nc!
> HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
>     CMD nc -z localhost 9009 || exit 1
> Do you want an actual HTTP /health endpoint? That might not be what [@schollz](https://github.com/schollz) wants. Reading through the source code, let us know [@felipewnp](https://github.com/felipewnp)!

The initial idea was an HTTP endpoint.

But since the application doesn't run on HTTP, but purely TCP (am I right?), it really wouldn't make sense to serve a healthcheck page over HTTP. 

The healthcheck with NC is a welcome addition though. 

I did something similar on kubernetes (example):

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: goproxy
  labels:
    app: goproxy
spec:
  containers:
  - name: goproxy
    image: registry.k8s.io/goproxy:0.1
    ports:
    - containerPort: 8080
    readinessProbe:
      tcpSocket:
        port: 8080
      initialDelaySeconds: 15
      periodSeconds: 10
    livenessProbe:
      tcpSocket:
        port: 8080
      initialDelaySeconds: 15
      periodSeconds: 10
```

If croc is purely TCP, please feel free to close this issue!

Issue #967, opened 2025-09-16
Docker image on docker hub not updated on release


dhjensen:
As title says. It would be nice to have an updated docker image again :)

Screenshot from the docker build process failing:

<img width="1522" height="688" alt="Image" src="https://github.com/user-attachments/assets/c985c597-b8d1-4772-8598-3ba19ec8b9a9" />

dhjensen:
I have not tested the new docker images but I can see they are available on docker hub on the new 10.2.7 release.
Closing.

Issue #946, opened 2025-08-06
[10.2.4]Send works well but resume not working


JerryJian:
Croc version: 10.2.4

_inernal.rar is a 2.2GB file. When I sent it using "send", the receiving end could receive it normally. However, the receiving end had no speed at 71.4% and was manually disconnected. When I sent it again, the receiving end attempted to resume the transfer but failed to connect consistently. Below are the detailed logs:

**Sender**
``` sender
croc.exe --debug send "_internal.rar"
[debug] 10:39:29 cli.go:215: debug mode on
[debug] 10:39:29 cli.go:219: public IP address: 60.21.107.78
[debug] 10:39:30 croc.go:531: hashed _internal.rar to dc0217bd9bf13942 using xxhash
[debug] 10:39:30 croc.go:536: file 0 info: {Name:_internal.rar FolderRemote:./ FolderSource:xxx Hash:[220 2 23 189 155 241 57 66] Size:2366697874 ModTime:2025-08-01 11:52:41.5675457 +0800 CST IsCompressed:false IsEncrypted:false Symlink: Mode:-rw-rw-rw- TempFile:false IsIgnored:false}
Sending 0 files (2.2 GB)[debug] 10:39:30 croc.go:540: longestFilename: 13
Sending '_internal.rar' (2.2 GB)
Code is: 4511-deal-engine-equator

On the other computer run:
(For Windows)
    croc 4511-deal-engine-equator
(For Linux/macOS)
    CROC_SECRET="4511-deal-engine-equator" croc
Code copied to clipboard
[debug] 10:39:30 croc.go:709: host: '168.119.238.198', port: '9009'
[debug] 10:39:30 croc.go:715: got host '168.119.238.198' and port '9009'
[debug] 10:39:30 croc.go:717: trying connection to 168.119.238.198:9009
[debug] 10:39:30 comm.go:78: dialing to 168.119.238.198:9009 with timelimit 5s
[debug] 10:39:30 tcp.go:88: starting with password 'p***3'
[info]  2025/08/06 10:39:30 starting TCP server on 0.0.0.0:9010
[debug] 10:39:30 tcp.go:88: starting with password 'p***3'
[info]  2025/08/06 10:39:30 starting TCP server on 0.0.0.0:9011
[debug] 10:39:30 tcp.go:88: starting with password 'p***3'
[debug] 10:39:30 tcp.go:88: starting with password 'p***3'
[...]

Issue #943, opened 2025-07-18
How to graceful cancel c.Send(...) and c.Receive() from go programm?


abakum:

schollz:
example?

abakum:
It would be great to see an example from the author!
```
go func() {
 c.Send(...)
 donechan <- true
}()
go func() {
 <-cancelchan
 ???
}()
...
```
```
go func() {
 c.Receive()
 donechan <- true
}()
go func() {
 <-cancelchan
 ???
}()
...
```

PR #969, opened 2025-09-26
feat: a better clipboard, a simple TCP healthcheck and fixed docker deployments

sansneo:
Hey, I love **croc** and I've been reading through the codebase today!

This PR should solve #967, provides a possible solution to #968, and makes #940 trivial to implement if @schollz wishes to do so. I personally think it's a neat idea!

Using `golang-design/clipboard` is a better way to handle clipboard support, as it covers every major system with every possible display server and clipboard provider.

The previous implementation worked fine but only covered Windows, macOS, and Linux (running Xorg with xclip or Wayland with wl-copy). Now clipboard support will work as expected on the BSDs and also on Linux systems using xsel instead of xclip.

I manually ran `golangci-lint` on the codebase and improved things here and there. 
Please review my other changes, these are minor and include some formatting improvements.

Using `gsa`(go size analyzer) on this I get an 11MB executable from:
```sh
CGO_ENABLED=0 go build -ldflags '-s -w -extldflags "-static"' -o croc
```
which might be a bit bloated. The whole codebase is pretty big.
```
┌─────────────────────────────────────────────────────────────────────────────┐
│ croc                                                                        │
├─────────┬──────────────────────────────────────────────┬────────┬───────────┤
│ PERCENT │ NAME                                         │ SIZE   │ TYPE      │
├─────────┼──────────────────────────────────────────────┼────────┼───────────┤
│ 18.27%  │ .rodata                                      │ 2.0 MB │ section   │
│ 15.33%  │ crypto                                       │ 1.6 MB │ std       │
│ 13.40%  │ net                                          │ 1.4 MB │ std       │
│ 8.27%   │ runtime                                      │ 886 kB │ std       │
│ 4.07%   │ .noptrdata                                   │ 436 kB │ section   │
│ 2.74%   │ reflect                                      │ 293 kB │ std       │
│ 2.30%   │ text/template                                │ 247 kB │ std       │
│ 2.25%   │ golang.org/x/net                             │ 241 kB │ vendor    │
│ 2.19%   │ github.com/schollz/croc/v10                  │ 235 kB │ main      │
│ 1.79%   │ math                                         │ 191 kB │ std       │
│ 1.69%   │ github.com/chzyer/readline                   │ 182 kB │ vendor    │
│ 1.64%   │ regexp                                       │ 176 kB │ std       │
│ 1.47%   │ slices                                       │ 158 kB │ std       │
│ 1.42%   │ time                                         │ 152 kB │ std       │
│ 1.38%   │ encoding/json                                │ 148 kB │ std       │
│ 1.34%   │ github.com/schollz/cli/v2                    │ 143 kB │ vendor    │
│ 0.00%   │ .text                                        │ 16 B   │ section   │
├─────────┼──────────────────────────────────────────────┼────────┴───────────┤
│ 100.00% │ Known                                        │ 11 MB              │
│ 100%    │ Total                                        │ 11 MB              │
└─────────┴──────────────────────────────────────────────┴────────────────────┘
```

Here are some additional notes and observations.

* The README is telling Alpine Linux users to install both bash and GNU coreutils to run this natively which is a big ask IMO. Maybe rewrite the installer in POSIX sh so Busybox ash can run it.
That said, using the Docker image works great!

* The installer checks $EUID and then tries to use `sudo` if that's not 0. This won't work anyone using `doas` over `sudo` or something else. Might be worth to tell `doas` users to install `doas-sudo-shim`!

* There are other tools with similar functionality to croc, such as `wormhole-william` that's also made in Go.

* Why -git over the expected standard --git and use a personal fork of `urfave/cli` instead of the upstream v3?

* While there aren't many environment variables used, you could use godotenv and a -e/--environment flag to point to a specific .env type file.

* A flag to specify the number of processors to use could be beneficial for performance tuning.

There's more!
#943 (goroutines) may be worth investigating.
#946 is a significant bug that should be addressed.

The codebase generally lacks comments and has a deeply nested structure (e.g., making changes requires updates in multiple places).

This was my attempt at contributing to croc.
I have a long way ahead in programming and hope I've been useful.
Thank you for the work you all put in!

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -45,5 +45,5 @@ jobs:
           file: ./Dockerfile
           platforms: linux/amd64,linux/arm,linux/arm64,linux/386
           push: ${{ github.event_name != 'pull_request' }}
-          tags: ${{ steps.docker_meta.outputs.tags }}
-          labels: ${{ steps.docker_meta.outputs.labels }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,9 @@
 go.work
 go.work.sum
 
-# env file
+# Environment variables file
 .env
+
+# Croc builds
+/croc
+croc_v*
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,5 +20,9 @@ COPY --from=builder /go/croc/croc /go/croc/croc-entrypoint.sh /
 
 USER nobody
 
+# Simple TCP health check with nc!
+HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 \
+    CMD nc -z localhost 9009 || exit 1
+
 ENTRYPOINT ["/croc-entrypoint.sh"]
 CMD ["relay"]
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/schollz/progressbar/v3 v3.18.0
 	github.com/skip2/go-qrcode v0.0.0-20200617195104-da1b6568686e
 	github.com/stretchr/testify v1.11.1
+	golang.design/x/clipboard v0.7.1
 	golang.org/x/crypto v0.42.0
 	golang.org/x/net v0.44.0
 	golang.org/x/sys v0.36.0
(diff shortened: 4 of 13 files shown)


schollz:
Sorry I had to undo this PR. This commit prevents `croc` from being statically compiled: 

```
$ CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -v -ldflags '-extldflags "-static"' 
```

sansneo:
Running the tests locally everything passes for me!
```
go test -v ./...
```

PR #970, opened 2025-09-27
feat: a better clipboard, a simple TCP healthcheck and fixed docker deployments

sansneo:
This should fix #969 so it can be merged.
It was undone as it failed a CI test and prevented the builds to be statically compiled.
My guess, it was a problem with the versioning as everything works locally for me.
The builds in ci.yml all compile statically for me and the tests pass.
See if this fixes it! (also synced)
A possible issue might be the Android/IOS clipboard as it might require an extra [dependency](https://github.com/golang-design/clipboard?tab=readme-ov-file#dependency), but I don't think that is the case!


sansneo:
Nope, working on it.

Issue #1000, opened 2025-11-18
--remember for relay


ThatOneShortGuy:
### What happened?

Looking at the code, there is not a remember option for `RelayAddress`.
```go
if errOpen == nil && !c.Bool("remember") {
	var rememberedOptions croc.Options
	err = json.Unmarshal(b, &rememberedOptions)
	if err != nil {
		log.Error(err)
		return
	}
	// update anything that isn't explicitly set
	if !c.IsSet("no-local") {
		crocOptions.DisableLocal = rememberedOptions.DisableLocal
	}
	if !c.IsSet("ports") && len(rememberedOptions.RelayPorts) > 0 {
		crocOptions.RelayPorts = rememberedOptions.RelayPorts
	}
	if !c.IsSet("code") {
		crocOptions.SharedSecret = rememberedOptions.SharedSecret
	}
	if !c.IsSet("pass") && rememberedOptions.RelayPassword != "" {
		crocOptions.RelayPassword = rememberedOptions.RelayPassword
	}
	if !c.IsSet("overwrite") {
		crocOptions.Overwrite = rememberedOptions.Overwrite
	}
	if !c.IsSet("curve") && rememberedOptions.Curve != "" {
		crocOptions.Curve = rememberedOptions.Curve
	}
	if !c.IsSet("local") {
		crocOptions.OnlyLocal = rememberedOptions.OnlyLocal
	}
	if !c.IsSet("hash") {
		crocOptions.HashAlgorithm = rememberedOptions.HashAlgorithm
	}
	if !c.IsSet("git") {
		crocOptions.GitIgnore = rememberedOptions.GitIgnore
	}
}
```

I am hosting my own relay so I can transfer my files on home network. I would prefer if they could all remember the relay so I wouldn't have to specify it every time.

Is there a reason why it isn't an available remembered option?

### What did you expect to happen?

Passing `--remember` should remember the `--relay` flag.

### Steps to reproduce

1. Send file using `--remember` flag and a custom relay `--relay`
2. Do it again without the `--relay` flag, will use default relay

### croc version

v10.2.7

### Operating System

Linux

### OS Version

Arch, Ubuntu 22.04, Windows 10

### Relevant log output

```shell

```

PR #1001, opened 2025-11-18
Add remember relay options (#1000)

ThatOneShortGuy:
A fix for #1000 

As someone who has never written Go before, that was pretty straight forward

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -356,11 +356,15 @@ func send(c *cli.Context) (err error) {
 		if !c.IsSet("git") {
 			crocOptions.GitIgnore = rememberedOptions.GitIgnore
 		}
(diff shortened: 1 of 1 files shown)


schollz:
@ThatOneShortGuy the reason they were taken out is because its annoying when the default server relay changes and the config files have to be redone (because it stores the old), so my one ask would be to fix this so that the remember relay option only works if the relay address is different than the default

ThatOneShortGuy:
@schollz I see what you're talking about. At this point my concern is when the default IP changes. We can't track that in code, and the existing configs already have something like `"RelayAddress": "46.62.254.13:9009"`.

The best solution to that I can think of is to add the default `models.DEFAULT_RELAY{6}` to the list of addresses to try (`[]string{c.Options.RelayAddress6, c.Options.RelayAddress}`). I didn't add that in this PR yet, wanted to get your opinion first.

PR #1076, opened 2026-03-07
PR: Federated Relay Pool Implementation 🚀

pablodz:
PROPOSAL.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.24-alpine AS builder
+FROM golang:1.25-alpine AS builder
 
 RUN apk add --no-cache git gcc musl-dev
 
--- a/FEDERATED_RELAY_POOL_PROPOSAL.md
+++ b/FEDERATED_RELAY_POOL_PROPOSAL.md
@@ -1,0 +1,699 @@
+# Federated Relay Pool for Croc - Architecture and Implementation Notes
+
+**Status:** Implemented in v10.5.0  
+**Last Updated:** 2026-03-09 | **Revision:** 2.2 (Implementation-aligned)
+
+---
+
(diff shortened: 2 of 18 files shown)


schollz:
Thank @pablodz , no rush, Please ping me with the results of your tests once finished!

pablodz:
@schollz, the proposal for a federated pool is on PROPOSAL.md that will we removed before merge if pass

schollz:
@pablodz 

Some thoughts, even before considering merging. I think I need buy-in from some community members on this. It is a big change. As the sole maintainer its even more complexity for me to maintain. That is all *fine* as long as this feature gets used. So I would ask that we have dedicated people providing nodes to use this before implementing it. I don't want to be in a situation where I am the only one providing nodes...its basically then back to where we started...except that I'm paying more (more servers) and I'm supporting more (problems with users creating issues related to the federatino).

pablodz:
Sure! Even if I donated 10 TB of unused server bandwidth, it’d still just be the two of us collaborating. Maybe building a SaaS around this is the way to go?

I'm picturing a managed 'Croc SaaS' for companies that move files bigger than 1 GB or burn through 1 TB+ of bandwidth. As a quick way to spread the word, the CLI could just print a short message or ad to stdout when people handle large files. You could also add enterprise-grade compliance so companies have the option to self-host or just use the managed service.

A lot of companies would rather just use a ready-to-go SaaS so they don't have to deal with deployment overhead, and it'd still be really cost-effective for large enterprises. It's a model that works well—just look at projects like FastAPI. It’s fully open-source, but they also offer managed cloud services and enterprise solutions right out of the box

schollz:
closing this, I don't see the community support necessary for a change like this.

PR #1097, opened 2026-05-06
fix: honor --exclude and --git when --zip is used

### LINKED ISSUE #1087: [Question]: Does zip option honor excludes or nested gitignores?
### What happened?

Does croc, when sending, honor the exclusisons defined with `--exclusions` or from nested gitignores defined with `--git` before zipping?

examples:
`croc send --exclude "node_modules,.venv,.terraform,.mypy_cache,.gradle" --zip large_folder_1`
`croc send --exclude "node_modules,.venv,.terraform,.mypy_cache,.gradle" --git --zip large_folder_1`
`croc send --git --zip large_folder_1`

### croc version

10.4.2

### Operating System

Windows

### OS Version

Windows 11 25H2 26200.8039

### Relevant log output

```shell

```

### Additional context

_No response_

costajohnt:
## Summary

Closes #1087.

In `--zip` mode, `croc send` builds the archive without consulting `--exclude` or the `--git` ignore set. The post-walk filter in `cli.go` runs after, but it can only match against the resulting `.zip` filename, never its contents, so excluded paths leak into the archive.

## Fix

Pass `ignoredPaths` and `exclusions` through to `utils.ZipDirectory` and skip matching entries during the walk, mirroring the non-zip filtering behaviour:

- `src/utils/utils.go::ZipDirectory` signature now takes `(destination, source string, ignoredPaths map[string]bool, exclusions []string)`. `--git` matches use `ignoredPaths[absPath]`; `--exclude` uses case-insensitive substring match against the zip-relative path, mirroring the existing post-filter in `cli.go`.
- `src/croc/croc.go:411` passes the existing `ignoredPaths` and `exclusions` through.

## Tests

Three new tests, all fail against `main`:

- `TestZipDirectoryHonoursExclusions` (utils): verifies `node_modules` and `.venv` are absent from a zip when given as `--exclude` patterns, and `main.go` is kept.
- `TestZipDirectoryHonoursIgnoredPaths` (utils): verifies a path in the `ignoredPaths` map is absent and the kept paths are present.
- `TestGetFilesInfoZipFolderHonoursFilters` (croc): end-to-end via `GetFilesInfo`. Populates a `.gitignore`, calls with both flags active, opens the resulting zip, and asserts the filtered contents.

`go test ./src/...`, `go build ./...`, and `go vet ./...` all clean locally.

## Note

The signature change to `utils.ZipDirectory` is technically an API break for any external consumer of `github.com/schollz/croc/v10/src/utils`. Only one in-tree caller exists. Happy to add a deprecated wrapper if you would rather not break it.

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -408,7 +408,7 @@ func GetFilesInfo(fnames []string, zipfolder bool, ignoreGit bool, exclusions []
 			}
 			fpath = filepath.Dir(fpath)
 			dest := filepath.Base(fpath) + ".zip"
-			utils.ZipDirectory(dest, fpath)
+			utils.ZipDirectory(dest, fpath, ignoredPaths, exclusions)
 			utils.MarkFileForRemoval(dest)
 			stat, errStat = os.Lstat(dest)
 			if errStat != nil {
--- a/src/croc/croc_test.go
+++ b/src/croc/croc_test.go
@@ -1,6 +1,7 @@
 package croc
 
 import (
+	"archive/zip"
 	"context"
 	"fmt"
 	"math/rand"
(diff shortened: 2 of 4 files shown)


schollz:
Looks great!

PR #1113, opened 2026-05-21
feat: skip file info transfer for /dev/null placeholder

abakum:


abakum:
## feat: skip file info transfer for /dev/null placeholder

**Problem**: In WebDAV/TCP forwarding mode, real files are transferred outside of croc — through a WebDAV server or TCP forwarding. However, croc's `Send()` method is the single entry point that performs several critical functions simultaneously:
1. PAKE key exchange — establishing a shared secret key
2. Channel encryption — creating an encrypted tunnel
3. File metadata transfer — information about files being sent
4. Actual data transfer — binary file stream

For tunnel-only use cases, only steps 1-3 are needed — no real file data should be sent through croc.

**Solution**: When the sender has exactly one file and it is `/dev/null`, skip the file info transfer by setting `Step2FileInfoTransferred = true` and returning early. The `/dev/null` placeholder:
- Exists in the filesystem — no errors when opening
- Has zero size — transfer completes instantly
- Contains no data — nothing extra is sent

**Analogy**: Like sending an empty envelope through a courier service so that:
- The courier (croc) establishes a secure route (encrypted channel)
- The recipient confirms their identity (PAKE authentication)
- After that, the real cargo travels via a separate highway (WebDAV/TCP forwarding)

This enables reusing croc's secure connection infrastructure (PAKE + encryption + authentication) as a transport tunnel for WebDAV or any other application, without sending any real files through the croc protocol itself.

PR #1115, opened 2026-05-21
feat: check relay password after both peers connect to room

abakum:


abakum:
## feat: check relay password after both peers connect to room

**Problem**: Currently, any client that does not know the relay password is rejected immediately. This means you cannot send files to a relay owner without knowing their password. A recipient who wants to receive a file from a relay owner's server must know the relay password.

**Solution**: Any peer can create a room on the relay (original behavior preserved). The authorization check is deferred to the moment when both peers have connected. If at least one peer knows the correct relay password, the transfer is allowed. If both peers are unauthorized (both using the default password `pass123`), the transfer is denied and the room is deleted.

**Benefits**:
- Allows sending files TO a relay owner without knowing their password — the sender connects with the default password, the relay owner connects with their correct password, one is authorized — transfer proceeds
- Prevents parasitic use: two strangers cannot use someone else's relay to transfer files between themselves
- Anti-spam: unauthorized rooms are automatically cleaned up after `UNAUTHORIZED_ROOM_TTL` (1 minute)
- Race condition fix: the "ok" notification is sent to the sender before piping starts, preventing a race where the sender starts sending before the pipe is established

**Tests**:
- `TestDefaultPasswordCanJoinExistingRoom`: sender with correct password, receiver with default → transfer works
- `TestCorrectPasswordFullAccess`: both peers with correct password → transfer works
- `TestArbitraryWrongPasswordRejected`: arbitrary wrong password (not default) → immediate rejection

**Implementation**:
- `roomInfo` struct gains `firstAuthorized bool` field
- Password validation allows default password to pass through; authorization is checked when second peer joins
- `UNAUTHORIZED_ROOM_TTL = time.Minute` constant for cleanup of unauthorized rooms
- Move `"ok"` send before `pipe()` call to prevent race condition

PR #1116, opened 2026-05-21
fix: skip flaky tests for Windows symlinks and receiver patience

abakum:


abakum:
## fix: skip flaky tests for Windows symlinks and receiver patience

### Fix 1: Skip symlink test on Windows

**Problem**: `TestCrocSymlink` fails on Windows because creating symlinks requires elevated privileges (Run as Administrator) or Developer Mode enabled. On CI running Windows, this test fails unless the runner has special configuration.

**Solution**: Add `t.Skip("skipping: symlinks require admin privileges on Windows")` at the beginning of the test when `runtime.GOOS == "windows"`.

### Fix 2: Skip receiver stdout test with invalid secret

**Problem**: After the relay room authorization changes (PR #1115), the receiver now waits patiently for the sender — up to one minute on someone else's relay, or three hours on their own. The test `TestReceiverStdoutWithInvalidSecret` expects the receiver to fail immediately with an invalid secret, which no longer happens.

**Solution**: Skip the test with a message explaining the behavioral change: "The receiver is now patient: it will wait for the sender for one minute on someone else's relay, or three hours on its own."

**Note**: This PR should be submitted after PR #1115 is merged, otherwise the skip reason for test 2 won't make sense.

PR #1156, opened 2026-07-08
fix: receive-code prompt input and EOF handling at prompts

camjac251:
## Summary

Three commits fixing interactive prompt input.

Pasting the code hangs the accept prompt. The "Enter receive code:" prompt used chzyer/readline, which reads stdin in raw mode through an internal buffer. Pasting the code followed by a `y` answer (or typing ahead) pulled the `y` into readline's buffer, and readline dropped it when the prompt returned, so `Accept (Y/n)` blocked forever on an answer the user already gave. readline also redraws the entire line on every keystroke, which shows up as backspace erasing the whole input on terminals where its width or cursor model is off (chzyer/readline#93; the project is unmaintained and the redraw rework in chzyer/readline#202 never merged). The prompt now reads a line in canonical mode like every other croc prompt: the tty driver handles echo and backspace, and type-ahead past the first newline stays in the kernel buffer for the next prompt. This removes the readline dependency and drops tab completion of the code words. The prompt also moves from stdout to stderr, matching every other croc prompt.

EOF at a yes-default prompt counts as consent. `GetInput` discarded the read error, so a closed or exhausted stdin came back as an empty string, indistinguishable from pressing enter. At the three prompts that default to yes, EOF silently consented: the receiver accepted and wrote incoming files, `--ask` approved the send, and the did-you-mean-send prompt sent the files. `GetInput` now returns the read error, and those prompts refuse when stdin cannot be read. Prompts that default to no already treat empty input as the safe answer. Scripted receivers that relied on a closed stdin hitting the yes-default should pass `--yes`; a script that runs `croc file.txt` with no stdin now gets an error pointing at `croc send` instead of exiting 0 without sending anything.

The prompts also share one buffered stdin reader now. Each `GetInput` call built a fresh reader and discarded it after one line, so on a pipe the first read could pull several lines into the buffer and later prompts lost them. `printf 'code\ny\n' | croc` worked before only because the ignored EOF fell into the yes-default; with the shared reader the piped `y` reaches the accept prompt as an explicit answer.

Why remove readline rather than replace it: the maintained fork ergochat/readline would fix the redraw bugs, but any raw-mode line editor reads stdin through its own buffer, so the pasted `y` would still be stranded there unless every later y/N prompt also went through the same editor instance. The prompt needs one line, once; the tty driver's canonical mode already provides echo and backspace on every platform with zero dependencies. If tab completion of the code words is worth keeping, I can redo this on top of ergochat/readline with the instance threaded through the later prompts.

## Test plan

- [x] `go build`, `go vet ./src/...`, and `go test ./src/cli/` pass (the `TestFindOpenPorts` failure in `src/utils` predates this change and depends on local ports being free)
- [x] PTY harness driving a real transfer through a local relay: pasting the code plus a `y` answer in one chunk previously hung at `Accept (Y/n)` and now completes the transfer
- [x] Same harness: typing the code normally, and typing a typo then correcting it with backspace, both still transfer
- [x] `printf 'code\ny\n' | croc` (piped, no tty) completes the transfer
- [x] A receiver with stdin from /dev/null refuses the transfer (peer sees "refusing files", nothing written to disk); before this change it accepted and wrote the file
- [x] `croc file.txt </dev/null` exits 1 with a clear message instead of silently exiting 0 without sending


camjac251:
@schollz possible to review?

schollz:
looks good!

PR #1166, opened 2026-07-15
fix: Restore Windows-on-ARM64 build to CI

### LINKED ISSUE #1086: [Bug]: Go 1.26 doesn't support Windows-on-ARM 32bit, so croc tests fail
### What happened?

Since you upgraded Go to 1.26 in CI tests, they fail to build croc on this line:
```
CGO_ENABLED=0 GOOS=windows GOARCH=arm go build -ldflags '-extldflags "-static"' -o croc-windows-arm.exe
```

CI tests to pass, at lease the build phase.

### Steps to reproduce

Run CI.

### croc version

not released yet

### Operating System

Windows

### OS Version

Check your CI

### Relevant log output

```shell
go version go1.26.1 linux/amd64
go: unsupported GOOS/GOARCH pair windows/arm
Error: Process completed with exit code 2.
```

### Additional context

This is because `GOOS=windows GOARCH=arm` support has been removed from Go in 1.26, as detailed [here](https://github.com/golang/go/issues/71671). `GOOS=windows GOARCH=arm64` is still supported, and croc still works with that.

Croc either needs to remove support for windows/arm 32bit, or continue to build windows/arm with Go 1.24 using a different stage.

It's notable that there were known issues with Go generated binaries on windows/arm 32bit, which may or may not have affected croc. We can't know without running the tests on windows/arm 32bit. I don't have a Windows RT device, and don't know how to run those tests otherwise, so I don't know how to run those tests. Windows 11 is dropping or has dropped support for running 32-bit ARM binaries, so you might have to search for used hardware to run these tests.

BrianHawley:
Go 1.26.0 removes Windows-on-ARM32 support, but keeps ARM64. I still use croc on Windows-on-ARM64 regularly, so I want to make sure that the target build keeps getting tested, so it won't be forgotten.

Fixes #1086

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,6 +28,7 @@ jobs:
         run: |
           go version
           CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -ldflags '-extldflags "-static"' -o croc-windows-amd64.exe
+          CGO_ENABLED=0 GOOS=windows GOARCH=arm64 go build -ldflags '-extldflags "-static"' -o croc-windows-arm64.exe
           CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags '-extldflags "-static"' -o croc-linux-amd64
           CGO_ENABLED=0 GOOS=linux GOARCH=386 go build -ldflags '-extldflags "-static"' -o croc-linux-386
           CGO_ENABLED=0 GOOS=linux GOARCH=arm go build -ldflags '-extldflags "-static"' -o croc-linux-arm


schollz:
@BrianHawley  can you check if its still on the release too?

BrianHawley:
@schollz for the release builds, Windows on ARM64, ARM32, and 32-bit x86 are still supported. The release builds are still built with Go 1.24.0. I wanted to add Windows on ARM64 back to CI so you wouldn't remove it from the release builds when you eventually upgrade Go to 1.26+ for release builds.

For this PR, I just copied the Windows-on-ARM64 line removed in https://github.com/schollz/croc/commit/1b7e260f61df0e10eca788d3593378233be74bcd and put it back where it was. I didn't add back the other unsupported platforms.

Windows on 32bit x86 is still supported by Go 1.26.x; Microsoft seems to have dropped support for it last year, but there are a lot of legacy systems. If you like, I can add that platform back too. I don't currently have a fully x86 32bit Windows platform installed anywhere, but we could set up a VM with it if necessary. Regular Windows 64bit and WINE can still run 32-bit x86 binaries though.

Issue #1163, opened 2026-07-11
[Feature Request]: Exclude files by absolute path or add explicit file exclusion


Ferase:
### What happened?

Currently, croc's `--exclude` flag only excludes substrings of the relative paths of the files/folders being sent. While unlikely, if a user would like to exclude a specific file in a specific directory from being sent, they can only pass the relative path. This can make it difficult (or impossible) to accurately exclude specific files in certain situations.

### What did you expect to happen?

I had thought croc would store each path absolutely, which would allow the `--exclude` flag to be used without the ambiguity of the relative path potentially playing havoc with other potential children of that path.

### Steps to reproduce

Here's an example that showcases the conundrum:

Say I have two files of the same name at two different locations under a root folder:
```
root_folder/subfolder_01/sub_subfolder_01/image.jpg
root_folder/subfolder_01/root_folder/subfolder_01/sub_subfolder_01/image.jpg
```
(yes i know these paths are highly nonsensical but hear me out)

The user would like to exclude the first `image.jpg`.

Thus, this will not work (it will exclude them both):
```
croc send --exclude "image.jpg" path/to/root_folder
```

Therefore, they need to put in the full relative path (because croc does not seem to consider paths absolutely):
```
croc send --exclude "root_folder/subfolder_01/sub_subfolder_01/image.jpg" path/to/root_folder
```

However, this will also exclude the second `image.jpg`, because:

**root_folder/subfolder_01/sub_subfolder_01/image.jpg** <- Matches in full just like they wanted
root_folder/subfolder_01/**root_folder/subfolder_01/sub_subfolder_01/image.jpg** <- Matches in part, but a substring match is still a match.

As far as I'm aware, this situation has no direct resolution without moving the unwanted `image.jpg` out of the folder or manually declaring the files you would like to send.

An edge case by all means, but still very much a possibility if someone has a child directory tree that somehow matches the root.

### croc version

croc version v10.4.13

### Operating System

Linux

### OS Version

EndeavourOS (rolling release)

### Relevant log output

```shell

```

### Additional context

The function of the `--exclude` flag is great for actual substring exclusion, but the purpose I am using it for (creating a [PyQt6 GUI for croc](https://github.com/Ferase/SwampSwap) with the intent of giving the user more granular control of the setup for sending files), it's way too broad and slightly out of scope for what the flag is meant to do.

[...]

PR #1174, opened 2026-07-21
Add exact file-path exclusion support via `--exclude-file`  #1163

SSameer20:
### Summary 
This change adds a new `--exclude-file` flag that excludes files by exact relative path, while
 preserving the existing `--exclude` substring/path-matching behavior.

 The goal is to let users exclude one specific file without accidentally excluding other files
 whose paths merely contain the same text.

### Problem (#1163)
The existing `--exclude` flag performs substring matching against relative file paths. This works well for broad exclusions (for example, excluding all files named image.jpg), but it cannot exclude only a single file when another file's path contains the same substring.

For example:
```
root_folder/
├── subfolder_01/
│   └── sub_subfolder_01/
│       └── image.jpg
└── subfolder_01/
    └── root_folder/
        └── subfolder_01/
            └── sub_subfolder_01/
                └── image.jpg
```

**Using:**
`croc send --exclude "root_folder/subfolder_01/sub_subfolder_01/image.jpg"`
causes both files to be excluded because the second file's relative path contains the first path as a substring.

### Solution

This PR introduces a new `--exclude-file` flag that performs exact normalized relative-path matching.

`--exclude` keeps its existing substring matching behavior for backward compatibility.
`--exclude-file` excludes only the file whose normalized relative path exactly matches the provided path.
The same matching logic is applied during both normal file discovery and archive (zip) creation.

### Validation

Verified manually using the directory structure from the reported issue.

### Results:

- No exclusions → 2 files transferred.
- `--exclude image.jpg` → both files excluded (existing behavior preserved).
- `--exclude-file subfolder_01/sub_subfolder_01/image.jpg` → only the targeted file is excluded, while the nested duplicate file is transferred.

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -79,6 +79,7 @@ func Run() (err error) {
 				&cli.IntFlag{Name: "transfers", Value: 4, Usage: "number of ports to use for transfers"},
 				&cli.BoolFlag{Name: "qrcode", Aliases: []string{"qr"}, Usage: "show receive code as a qrcode"},
 				&cli.StringFlag{Name: "exclude", Value: "", Usage: "exclude files if they contain any of the comma separated strings"},
+				&cli.StringFlag{Name: "exclude-file", Value: "", Usage: "exclude files matching any of the comma separated relative paths exactly"},
 				&cli.StringFlag{Name: "socks5", Value: "", Usage: "add a socks5 proxy", EnvVars: []string{"SOCKS5_PROXY"}},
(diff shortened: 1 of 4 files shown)


schollz:
@SSameer20  can you resolve conflicts? otherwise looks good.

SSameer20:
> @SSameer20 can you resolve conflicts? otherwise looks good.
Thanks! I noticed the PR has already been merged. Appreciate the review and feedback!

PR #1179, opened 2026-07-22
docs: document --exclude-file flag

sridhar-3009:
## Context

Related to #1163.

\`--exclude-file\` already exists in the CLI (\`src/cli/cli.go\`) and does exact relative-path matching via \`exactPathExcluded\`, as opposed to \`--exclude\`'s substring matching. It just isn't mentioned anywhere in the README, so users hit the substring-collision problem described in #1163 without knowing there's already a flag that solves it.

## Change

Adds a short explanation to the "Excluding Folders" section distinguishing the two flags, with an example for each.

I left a comment on #1163 pointing to \`--exclude-file\` as the existing fix — this PR closes the documentation gap that made the issue possible in the first place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,10 +223,16 @@ croc --yes --overwrite <code>
 
 #### Excluding Folders
 
-To exclude folders from being sent, use the `--exclude` flag with comma-delimited exclusions:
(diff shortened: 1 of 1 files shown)


schollz:
thank you!!!

PR #1183, opened 2026-07-23
feat: add --rename flag to receive existing files under a new name

### LINKED ISSUE #1182: [Feature Request]: Option to keep both files instead of overwrite/skip when a file already exists
### What happened?

A friend sent me `video.mkv`, but I already had a different file with the same name in my download folder. When I ran `croc <code>`, the only choice I got was:

```
Overwrite 'video.mkv'? (y/N) (use --overwrite to omit)
```

So my options were to overwrite my existing file or skip the transfer entirely. To receive the file without losing my own copy, I had to cancel, rename my local file manually, and run the transfer again.

An option to keep both files, saving the incoming one under a new name like `video (1).mkv`, the way browsers and most file managers handle name collisions. This could be either:

- an extra choice at the prompt, e.g. `Overwrite 'video.mkv'? (y/N/r)` where `r` renames the incoming file, or
- a flag like `--rename` that auto-renames on conflict without prompting (similar to `wget`'s `file.1` behavior)

I understand this interacts with the resume feature, since a same-name file with a different hash is currently treated as a partial transfer to resume.

I'd like to implement this myself, so please let me know which approach you prefer and I'll open a PR.

xevrion:
### Summary

This adds a `--rename` flag for receivers. When an incoming file already exists locally with different content, croc saves it as `name (1).ext` (incrementing until unused) instead of prompting to overwrite or resume. Default behavior without the flag is unchanged, as requested in #1182.

### Problem (#1182)

When receiving a file whose name already exists locally, the only choices today are overwriting the local file or skipping the transfer. To keep both files the receiver has to cancel, rename their local copy by hand, and run the transfer again.

### Solution

With `--rename`, the receiver picks the first unused `name (N).ext` in the destination folder and receives into that, printing `Receiving 'video.mkv' as 'video (1).mkv'`. If the existing file is identical to the incoming one, the transfer is skipped as before, so no duplicate is created. The flag also works with `--remember`.

Verified with the new unit test for the name picker, a new integration test (`TestCrocRenameExistingFile`), and manual transfers covering: conflict with `--rename` (new file created, original untouched), `video (1).mkv` also taken (increments to `(2)`), identical file present (skipped, no duplicate), no flags (prompt unchanged), and `--overwrite` (still overwrites in place).

Fixes #1182.

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,14 @@ To automatically overwrite files without prompting, use the `--overwrite` flag:
 croc --yes --overwrite <code>
 ```
 
+#### Keep Both Files Without Prompt
+
+To keep an existing file and receive the incoming one under a new name (e.g. `video (1).mkv`), use the `--rename` flag:
+
+```bash
+croc --yes --rename <code>
+```
+
 #### Excluding Folders
 
 To exclude folders from being sent, use the `--exclude` flag with comma-delimited exclusions:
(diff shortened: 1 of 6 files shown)


schollz:
Looks good.

PR #1188, opened 2026-07-24
Fix/web receiver silent hang

NICE-DEV226:
Quick note for review 👇

Heads-up that the web client tests don't run in CI (the workflow only does
`go test ./...`), so these two fixes ship with their own Vitest coverage you can
run locally:

The hang was the one that bit me: if a data socket drops before the first file
is requested, `receiveFiles()` used to wait on a promise that never settled —
and the "Cancel receive" button is hidden at that point, so the only way out was
a page reload. The new test times out against `main` and passes with the fix,
which is the clearest way to see it.

Both changes are independent — happy to split them into two PRs if you'd prefer
to review/merge them separately. And I have a couple more small fixes queued for
the fresh web code (a nil-deref guard on `os.Stdin.Stat()`, a WASM handle leak on
cancelled transfers) that I'll send as separate PRs so this one stays focused.

Thanks for croc — the WASM-reuse approach for the browser client is really nice.

### DIFF
--- a/web/src/protocol/client.test.ts
+++ b/web/src/protocol/client.test.ts
@@ -1,0 +1,74 @@
+import { describe, expect, it, vi } from "vitest";
+
+// The receiver spins up a WASM worker to decrypt data; none of these tests
+// reach a decrypt, so a stub keeps them off the (jsdom-less) Worker path.
+vi.mock("../wasm/client", () => ({
+  wasm: () => ({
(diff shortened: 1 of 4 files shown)


schollz:
Looks great, thanks!

And yeah, I haven't added it to CI because my Github Actions are already so expensive :( but I do check things locally.

PR #1190, opened 2026-07-25
fix: return error instead of panic in EncryptChaCha

skdishansachin:
If `crypto/rand` fails during nonce generation, `EncryptChaCha` currently panics instead of returning the error. The function already returns `error`, so this changes the panic to `return nil, err`.


schollz:
thank you!

PR #1191, opened 2026-07-25
fix: reject receive paths through symlink parents

divyanshu-iitian:
## Summary

- reject existing symlinks in every destination path component, not only the final filename
- apply the guard to regular files, empty files/folders, transferred symlinks, and automatic ZIP extraction
- pre-validate ZIP entries before extraction and re-check immediately before writes
- add regression coverage for direct receives and ZIP extraction through symlinked parent directories

## Why

The receiver already rejected a final destination that was a symlink, but an existing parent symlink was still followed. For example, if `nested` pointed outside the receive directory, a peer-provided `nested/payload.txt` could be written outside that directory. Automatic ZIP extraction had the same gap.

The shared guard resolves the target relative to the intended root and walks each existing component with `Lstat`, rejecting symlinks before any write.

## Validation

- `go vet ./...`
- focused receive/unzip security tests pass
- all non-`src/croc` packages pass in `go test ./...`
- the full Windows run reaches the existing `TestCrocSymlink` privilege failure (`A required privilege is not held by the client`); Linux CI exercises the new symlink regression tests

### DIFF
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -712,17 +712,7 @@ func validateReceiveMetadata(files []FileInfo, emptyFolders []FileInfo) ([]FileI
 }
 
 func rejectSymlinkDestination(pathToFile string) error {
-	info, err := os.Lstat(pathToFile)
-	if os.IsNotExist(err) {
-		return nil
-	}
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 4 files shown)


schollz:
thank you!

PR #1192, opened 2026-07-26
Fix/web receiver silent hang

NICE-DEV226:


schollz:
looks good! thank you!!

PR #1236, opened 2026-08-11
Add NoEncrypt option to CLI and configuration structures

### LINKED ISSUE #1222: [Feature request]: option to disable encryption
### What happened?

Hey!

One of my main use cases with `croc` is using it over a trusted LAN network or p2p ethernet. I believe it would be helpful to get rid of the encryption overhead, especially with faster networking speeds becoming more common and there are scenarios where every minute/second matters during a file transfer (some LAN computer competitions for example)

You may say I should use another program then, but the easy to use and explain, and resumbability features of croc make it something that I haven't been easily able to replace

Please consider this!

(ps, it is impossible to make a non bug report issue after commit c811a6e91bbaae677760189825d0028c2f1eb7e0)

N/A

### Steps to reproduce

N/A

### croc version

N/A

### Operating System

Linux

### OS Version

N/A

### Relevant log output

```shell

```

### Additional context

_No response_

arungosukula:
Fixes #1222 
## Summary

Added an optional --no-encrypt mode to croc for file transfers over trusted networks such as LAN or direct P2P Ethernet connections.

## Motivation

For high-speed trusted network transfers, payload encryption can introduce additional CPU overhead and may reduce transfer performance.

croc already provides useful features such as simple peer-to-peer transfer and resumable file transfers. This change allows users to retain those features while optionally disabling payload encryption when encryption is not required.

## Changes

* Added a new --no-encrypt CLI option.
* Added NoEncrypt to the transfer options.
* Added encryption mode information to SenderInfo so the receiver can determine whether payload encryption is enabled.
* Updated the sender to skip payload encryption when --no-encrypt is enabled.
* Updated the receiver to skip payload decryption for unencrypted transfers.
* Kept the existing PAKE/session setup unchanged.
* Kept encrypted transfers as the default behavior.
* Preserved the existing compression and resumable transfer functionality.
* Added/updated tests for encrypted and unencrypted transfer scenarios.
* Updated documentation for the new option.

## Usage

Existing encrypted transfer remains unchanged:

bash
croc send <file>

For a trusted network:

bash
croc send --no-encrypt <file>

The receiver does not need to specify an additional option. The encryption mode is communicated by the sender as part of the transfer metadata.

## Security Consideration

--no-encrypt is explicitly opt-in and should only be used on trusted networks where payload confidentiality is not required.

The default behavior remains unchanged, with payload encryption enabled.

## Validation

Validated the following scenarios:

* Normal encrypted file transfer
* Unencrypted file transfer
* Unencrypted transfer with compression enabled
* Unencrypted transfer with compression disabled
* Resumable unencrypted transfer
* File integrity after transfer
* Existing encrypted transfer behavior remains unchanged

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -76,6 +76,7 @@ func newApp() *cli.App {
 				&cli.StringFlag{Name: "text", Aliases: []string{"t"}, Usage: "send some text"},
 				&cli.BoolFlag{Name: "no-local", Usage: "disable local relay when sending"},
 				&cli.BoolFlag{Name: "no-multi", Usage: "disable multiplexing"},
+				&cli.BoolFlag{Name: "no-encrypt", Usage: "disable payload encryption for trusted networks"},
 				&cli.BoolFlag{Name: "git", Usage: "enable .gitignore respect / don't send ignored files"},
 				&cli.IntFlag{Name: "port", Value: 9009, Usage: "base port for the relay"},
 				&cli.IntFlag{Name: "transfers", Value: 4, Usage: "number of ports to use for transfers"},
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -381,6 +381,7 @@ func send(c *cli.Context) (err error) {
 		RelayPorts:        ports,
 		Ask:               c.Bool("ask"),
 		NoMultiplexing:    c.Bool("no-multi"),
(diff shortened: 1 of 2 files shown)


schollz:
@arungosukula failing checks

arungosukula:
Hi @schollz,

The code has been fixed and the latest changes are working as expected. I have successfully tested --no-encrypt file transfer between two systems using the newly built binary.

PR #1237, opened 2026-08-11
feat: add --json machine-readable event output

### LINKED ISSUE #1201: Feature request: add machine-readable JSON event output
## Feature request

Add a machine-readable output mode for wrappers and automation, e.g. `--json` or `--events json`.

## Motivation

GUI wrappers, scripts, and higher-level tools currently have to parse human-oriented stdout/stderr to detect:

- generated code phrase
- current phase: hashing / connecting / securing channel / transferring / verifying / complete
- file name, total bytes, transferred bytes, speed, percent
- prompts / confirmations
- stored-transfer link, token, expiry, and revoke id
- structured errors

That makes wrappers fragile when text changes. A stable event stream would let third-party clients build on croc without scraping terminal UI output.

## Proposed behavior

One additive option, preserving current output by default:

```bash
croc --events json send file.zip
# or
croc --json send file.zip
```

Emit newline-delimited JSON events to stdout or stderr, one object per state change:

```json
{"type":"version","version":"10.7.0"}
{"type":"code","code":"alpha-forest-42"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
{"type":"progress","file":"file.zip","bytesTransferred":1048576,"bytesTotal":10485760,"percent":10.0,"speedBps":524288}
{"type":"complete","files":[{"name":"file.zip","bytes":10485760}]}
```

For stored transfers:

```json
{"type":"store","url":"https://getcroc.com/s/...#v1...","token":"croc-store-v1...","expiresAt":"2026-07-30T03:00:00Z","revokeId":"..."}
```

For errors:

```json
{"type":"error","code":"relay_unreachable","message":"could not connect to relay","hint":"check --relay or proxy settings"}
```

## Design notes

- Keep human output unchanged unless the flag is set.
- Prefer newline-delimited JSON over one final JSON blob so long transfers can stream progress.
- Avoid secrets in process args where possible; current `CROC_SECRET` / `CROC_STORE_TOKEN` env patterns are still useful.
- If both human and JSON output are desired, an explicit `--events json --quiet=false` style split could be discussed.

## Why this helps croc itself

This makes croc easier to embed in:

- desktop/mobile GUIs
- backup scripts
- CI jobs
- chat/agent workflows
- troubleshooting tools like a future `croc doctor`

It also reduces downstream bug reports caused by wrappers parsing changed human text.

MeGaurav4:
## What

Adds a global `--json` flag that makes croc emit newline-delimited JSON events on stderr instead of the human-oriented UI. The default output is completely unchanged.

```
$ croc --json send bigfile.bin  2> events.ndjson
{"type":"version","version":"11.0.3"}
{"type":"code","code":"secret-phrase"}
{"type":"phase","phase":"hashing","message":"hashing files"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
{"type":"phase","phase":"transferring","message":"transferring files"}
{"type":"progress","file":"bigfile.bin","bytesTransferred":...,"bytesTotal":...,"percent":...,"speedBps":...}
{"type":"complete","files":[{"name":"bigfile.bin","bytes":...}]}
{"type":"phase","phase":"complete","message":"transfer complete"}
```

## Event types

| type | sender | receiver | payload |
|---|---|---|---|
| `version` | yes | yes | croc version |
| `code` | yes | - | code phrase the recipient needs |
| `phase` | yes | yes | hashing / connecting / transferring / complete |
| `progress` | yes | yes | file, bytesTransferred, bytesTotal, percent, speedBps (throttled to 100 ms) |
| `complete` | yes | yes | list of transferred files with sizes |
| `store` | stored-send only | - | browser URL, token, expiry, share ID |
| `error` | both | both | error message with stable code (on non-zero exit) |

## Implementation

- New `src/events` package: off by default, `events.Enable(os.Stderr)` opts in. All existing code paths and tests are unaffected.
- `--json` suppresses progress bars, status lines, the QR code, clipboard copy, and hashing bars so the stream stays machine-parseable (verified end-to-end).
- `complete` and `transferring` are emitted exactly once per process.
- Wrappers can read the code phrase before any recipient connects, and exit with a non-zero status on error.

## Testing

- `src/events`: event encoding tests.
- `src/croc`: end-to-end tests covering direct transfers (sender + receiver event sequences) and stored transfers (`store` + `complete` events).
- Live relay smoke test: sender/receiver streams parse as pure NDJSON with zero human noise, transfer byte-identical.

Closes #1201

### DIFF
--- a/src/cli/cli.go
+++ b/src/cli/cli.go
@@ -17,6 +17,8 @@ import (
 	"github.com/schollz/croc/v11/src/codephrase"
 	"github.com/schollz/croc/v11/src/comm"
 	"github.com/schollz/croc/v11/src/croc"
+	"github.com/schollz/croc/v11/src/events"
+	"github.com/schollz/croc/v11/src/events"
 	"github.com/schollz/croc/v11/src/models"
 	"github.com/schollz/croc/v11/src/storeclient"
 	"github.com/schollz/croc/v11/src/tcp"
@@ -34,7 +36,11 @@ func Run() (err error) {
 	// use all of the processors
 	runtime.GOMAXPROCS(runtime.NumCPU())
 
-	return newApp().Run(os.Args)
+	err = newApp().Run(os.Args)
+	if err != nil && events.Enabled() {
+		events.Error(err)
+	}
+	return
 }
 
 func newApp() *cli.App {
@@ -137,6 +143,7 @@ func newApp() *cli.App {
 		&cli.BoolFlag{Name: "rename", Usage: "receive files that already exist under a new name instead of prompting"},
 		&cli.BoolFlag{Name: "testing", Usage: "flag for testing purposes"},
 		&cli.BoolFlag{Name: "quiet", Usage: "disable all output"},
+		&cli.BoolFlag{Name: "json", Usage: "output machine-readable JSON events to stderr"},
 		&cli.BoolFlag{Name: "disable-clipboard", Usage: "disable copy to clipboard"},
 		&cli.BoolFlag{Name: "extended-clipboard", Usage: "copy full command with secret as env variable to clipboard"},
 		&cli.StringFlag{Name: "revoke", Usage: "revoke a stored transfer using its local sender receipt"},
@@ -399,6 +406,7 @@ func send(c *cli.Context) (err error) {
 		Quiet:             c.Bool("quiet"),
 		DisableClipboard:  c.Bool("disable-clipboard"),
 		ExtendedClipboard: c.Bool("extended-clipboard"),
+		Events:            c.Bool("json"),
 	}
 	if crocOptions.RelayAddress != models.DEFAULT_RELAY {
 		crocOptions.RelayAddress6 = ""
@@ -567,6 +575,11 @@ Or you can go back to the classic croc behavior by enabling classic mode:
 		return
 	}
 
+	if c.Bool("json") {
+		events.Version(Version)
+		events.Code(crocOptions.SharedSecret)
+	}
+
 	// save the config
 	saveConfig(c, crocOptions)
 	err = cr.Send(minimalFileInfos, emptyFoldersToTransfer, totalNumberFolders)
@@ -693,6 +706,7 @@ func receive(c *cli.Context) (err error) {
 		Quiet:             c.Bool("quiet"),
 		DisableClipboard:  c.Bool("disable-clipboard"),
 		ExtendedClipboard: c.Bool("extended-clipboard"),
+		Events:            c.Bool("json"),
 	}
 	if crocOptions.RelayAddress != models.DEFAULT_RELAY {
 		crocOptions.RelayAddress6 = ""
@@ -821,6 +835,10 @@ Or you can go back to the classic croc behavior by enabling classic mode:
 	cr, err := croc.New(crocOptions)
 	if err != nil {
 		return
+	}
+
+	if c.Bool("json") {
+		events.Version(Version)
 	}
 
 	// save the config
--- a/src/cli/store.go
+++ b/src/cli/store.go
@@ -15,6 +15,8 @@ import (
 	"github.com/rivo/uniseg"
 	"github.com/schollz/croc/v11/internal/cli"
 	"github.com/schollz/croc/v11/src/croc"
+	"github.com/schollz/croc/v11/src/events"
+	"github.com/schollz/croc/v11/src/events"
 	storeapi "github.com/schollz/croc/v11/src/store"
 	"github.com/schollz/croc/v11/src/storeclient"
 	"github.com/schollz/croc/v11/src/storecrypto"
@@ -247,9 +249,12 @@ func sendStored(c *cli.Context) error {
 		strings.TrimSpace(c.String("store-url")),
 		paths,
 		storeclient.UploadOptions{Downloads: downloads, Expiration: expiration},
-		storedCallbacks(c.Bool("quiet")),
+		storedCallbacks(c.Bool("quiet") && !c.Bool("json")),
 	)
-	if !c.Bool("quiet") {
+	if c.Bool("json") {
+		events.Version(Version)
+	}
+	if !c.Bool("quiet") && !c.Bool("json") {
 		fmt.Fprintln(os.Stderr)
 	}
 	if err != nil {
@@ -270,6 +275,15 @@ func sendStored(c *cli.Context) error {
 		ExpiresAt:   result.ExpiresAt,
 	}); err != nil {
 		return fmt.Errorf("save stored-transfer revoke receipt: %w", err)
+	}
+	if c.Bool("json") {
+		events.Store(
+			browserURL,
+			token,
+			result.ExpiresAt.UTC().Format(time.RFC3339),
+			result.Share.ID,
+		)
+		return nil
 	}
 	downloadLimit := fmt.Sprintf("%d verified downloads", result.Downloads)
 	if result.Downloads == 1 {
--- a/src/croc/croc.go
+++ b/src/croc/croc.go
@@ -38,6 +38,8 @@ import (
 	"github.com/schollz/croc/v11/src/comm"
 	"github.com/schollz/croc/v11/src/compress"
 	"github.com/schollz/croc/v11/src/crypt"
+	"github.com/schollz/croc/v11/src/events"
+	"github.com/schollz/croc/v11/src/events"
 	"github.com/schollz/croc/v11/src/message"
 	"github.com/schollz/croc/v11/src/models"
 	"github.com/schollz/croc/v11/src/pakekey"
@@ -109,6 +111,7 @@ type Options struct {
 	Quiet             bool
 	DisableClipboard  bool
 	ExtendedClipboard bool
+	Events            bool
 }
 
 type SimpleMessage struct {
@@ -173,6 +176,7 @@ type Client struct {
 	senderRouteReady        chan struct{}
 	senderRouteReadyOnce    sync.Once
 	transferStarted         atomic.Bool
+	transferComplete        atomic.Bool
 	// localRelayPort is the control port of the ephemeral local relay started by
 	// setupLocalRelay(). It is captured before any goroutines that might
 	// overwrite c.Options.RelayPorts are launched.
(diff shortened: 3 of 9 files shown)


schollz:
thanks @MeGaurav4 please have @interfluve-wav review it

interfluve-wav:
## Review from `interfluve-wav`

Thanks @MeGaurav4 — we (Croc-GUI maintainers) have skin in this since our Tauri
backend currently parses croc's stderr with a custom regex-based
`parse_progress_line`, and a stable `--json` mode would let us delete a lot of
that. So this is exactly the shape we need. A few notes from a live test.

### Verified locally
- `go test ./src/events/ -race -v` → 3/3 pass.
- `go test ./src/croc/ -run TestJSONEventsE2E -v` → pass (without `-race`).
- Built `croc` with the PR branch and ran a real `send` + `receive` over loopback
  with `--json`. Both stderr streams are **pure NDJSON** (every line is valid
  JSON with a `type` field, zero human noise). That's the contract we need.

Sample sender stream from my run:
```json
{"type":"version","version":"11.1.0"}
{"type":"code","code":"jsontest-…"}
{"type":"phase","phase":"hashing","message":"hashing files"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
```

Receiver stream:
```json
{"type":"version","version":"11.1.0"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
```

Looks great. Now the asks:

### 🔴 Worth fixing before merge

**1. Error classifier misses the most common failure mode.**
Running send + receive where the receiver doesn't accept the offered file list,
both sides emit `{"type":"error","code":"failed","message":"refused files"}` and
`"peer error: refusing files"`. Both land in the catch-all `"failed"` bucket, so
a wrapper can't distinguish *refused* from *unreachable* from *auth*.
`classify()` in `src/events/events.go` only matches `"could not connect"`,
`"bad password"`, `"timed out"`, and `"could not secure channel"`. Suggest:
- Add a case for `"refusing files"` / `"refused files"` → `code: "files_refused"`,
  `hint: "the recipient declined the offered files"`.
- Add a case for `"peer error"` → `code: "peer_error"`.
- The `"passphrase"` substring is broad — it could match an unrelated warning in
  future error text. Prefer matching `relay password` / `bad password` exactly
  (which the code already handles), and consider dropping the bare `passphrase`
  match unless you have a specific error string in mind.

**2. `Error` event isn't emitted from the CLI exit path consistently.**
`src/cli/cli.go` does:
```go
err = newApp().Run(os.Args)
if err != nil && events.Enabled() {
    events.Error(err)
}
return
```
That's correct. But `croc send` and `croc receive` can also exit non-zero from
inside their handler without returning through this wrapper (e.g. when
`c.Send()` / `c.Receive()` returns an error and the handler returns it). In
those cases the error already propagates, so `events.Error(err)` does fire — but
the test would be worth adding. Consider an e2e test that runs `croc --json
receive` with a bad code phrase and asserts an `error` event lands on stderr
with a non-`"failed"` code.

### 🟡 Suggestions (non-blocking)

**3. The e2e test only asserts `connecting` and `transferring` phases are present.**
It does not assert that `complete` lands at the end. For a wrapper, missing
`complete` is the difference between "transfer done" and "stuck at 100%". Add:
```go
if !sawComplete {
    t.Error("expected complete event")
}
```
(There's a similar soft check on `len(complete.Files) != 2` that doesn't gate
test failure unless combined with `!sawComplete` — make the test fail loud.)

**4. `version` field on the receiver.**
The receiver emits `version` on startup, before the handshake. That's fine, but
it means a wrapper reading both streams will see *two* `version` events for one
transfer. Either:
- Add a `role: "sender" | "receiver"` field to `VersionEvent`, or
- Document that downstream consumers should dedupe by `version` and treat it
  as per-process.

**5. `humanOutputMuted` is process-global.**
If `events.Enable(os.Stderr)` is called from a test, then `MuteHumanOutput(true)`
is set globally and isn't reset by `events.Enable(nil)`. A
`defer termui.MuteHumanOutput(false)` in the test would fix this; otherwise
test ordering matters. Not a production concern.

**6. Token in `store` event.**
`Store` already includes the download token, which is sensitive (anyone with
it can fetch the file). This is a **pre-existing** surface — the non-JSON
`formatStoredInstructions` template also prints it — but the structured form is
easier to grep and to accidentally log. A `secret: true` flag on the field, or
a separate `--json` opt-in for the store token (`--json-include-secrets`),
would be a safer default. Not blocking — this is a policy question for
@schollz.

**7. `MuteHumanOutput` for `--json` mode in stored transfers.**
In `src/cli/store.go`, `storedCallbacks(quiet && !json)` is correctly throttled
when `--json` is set. But the early `fmt.Fprintln(os.Stderr)` on the trailing
blank line is gated with `!quiet && !json` — good. Just confirming the stored
path was consistent with the rest.

### Looks good
- `atomic.Bool` for `transferComplete` correctly guards double-emit of `complete`.
- `termui.MuteHumanOutput` is the right choke point — kills the human UI
  without touching every call site.
- 100ms progress throttle is sane.
- `omitempty` on optional `StoreEvent` / `ErrorEvent` fields is correct.
- Stderr (not stdout) for events is the right choice — keeps the file content
  stream clean.

### Verdict
**Concept: ship it. Implementation: needs the classifier fix and one extra
assertion in the e2e test before merge.** Happy to open a follow-up PR against
this branch with the classifier additions if @MeGaurav4 wants to keep this PR
small.

For our side: Croc-GUI will adopt `--json` as the default in the next minor.
I'll cut a follow-up issue in `interfluve-wav/croc-gui` to track the migration
once this lands.

interfluve-wav:
## Follow-up: corrected evidence (was wrong about the classifier)

@MeGaurav4 @schollz — I owe you a correction. My earlier review flagged the
error classifier as "missing `refused files`" because my local test harness
was hitting that path on every transfer. I just traced it and the harness
had bad flag positions (`--no-local` after `send` instead of before, and
`CROC_SECRET=***` which is too short). Once I fixed the harness, the PR
behaves correctly.

### What I got wrong in the first review

- **"Classifier misses `refused files`"** — the `"refused files"` error
  string only appears when something aborts *after* PAKE but *before*
  file confirmation (e.g. receiver has no `--ignore-stdin` and the parent
  process kills stdin). That's a generic "abort" signal, not a real auth
  / network failure. It's reasonable for it to land in the `failed`
  bucket. **I retract that ask.**
- **"Stored-transfer token in `Store` event is a new exposure"** — same
  flag-ordering issue was making me think transfers were completing when
  they weren't, which made me misread the stored path. The token was
  already in human output; the structured form is fine.

### What still holds from the first review

- ✅ The `complete` e2e test assertion is too soft
  (`if !sawComplete || len(complete.Files) != 2` — `sawComplete` is
  never gated as a hard `t.Error`). Worth tightening.
- ✅ `version` event fires on both sides, so a wrapper reading both
  streams sees two `version` events per transfer. Either add
  `role: "sender" | "receiver"` to `VersionEvent`, or document dedupe.
- ✅ `humanOutputMuted` is process-global; a `Reset()` helper or
  per-Client scope would be safer for tests. Not a production concern.
- ⚠️ The token in the `Store` event is sensitive (anyone with it can
  download). Pre-existing surface, but the structured form is more
  greppable. Worth a `secret: true` flag or a separate
  `--json-include-secrets` opt-in. **Policy question for @schollz, not
  blocking.**

### Verified end-to-end with the public relay

Sender:
```bash
croc --yes --json send --no-local /path/to/file
```

Receiver:
```bash
croc --yes --overwrite --ignore-stdin --json receive
```

Both sides emit clean NDJSON streams. Sample sender run, 80B file:
```json
{"type":"version","version":"11.1.0"}
{"type":"code","code":"finalfix1786941183"}
{"type":"phase","phase":"hashing","message":"hashing files"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
{"type":"phase","phase":"transferring","message":"transferring files"}
{"type":"progress","file":"test.txt","bytesTransferred":80,"bytesTotal":80,"percent":100,"speedBps":0}
{"type":"complete","files":[{"name":"test.txt","bytes":80}]}
{"type":"phase","phase":"complete","message":"transfer complete"}
```

Receiver (no `code` event, since the receiver never sees the secret):
```json
{"type":"version","version":"11.1.0"}
{"type":"phase","phase":"connecting","message":"connecting to relay"}
{"type":"phase","phase":"transferring","message":"transferring files"}
{"type":"progress","file":"test.txt","bytesTransferred":80,"bytesTotal":80,"percent":100,"speedBps":0}
{"type":"complete","files":[{"name":"test.txt","bytes":80}]}
{"type":"phase","phase":"complete","message":"transfer complete"}
```

Both processes exit 0. **0 invalid JSON lines** across both streams.

### Error path also verified (wrong code phrase on receiver)

Sender emits:
```json
{"type":"error","code":"auth_failed","message":"cipher: message authentication failed","hint":"make sure both sides use the same code phrase and relay password"}
```

Receiver emits:
```json
{"type":"error","code":"handshake_failed","message":"could not secure channel","hint":"the code phrase may be wrong, or the relay may be unreachable"}
```

So the classifier **does** work for the two most common failure modes
(auth + handshake). The catch-all `failed` only catches the obscure
"abort" path, which is fine.

### Bottom line

The PR is in good shape. The only thing I'd still want before merge is
the hard `complete` assertion in the e2e test. For our side
(`interfluve-wav/croc-gui`), this is ready to integrate — we can replace
~400 lines of regex-based progress parsing with ~50 lines of
`serde_json` deserialization. I'll open a follow-up issue in
`interfluve-wav/croc-gui` once this lands.

Sorry for the noise on the first pass.

MeGaurav4:
Thanks for the thorough review and the correction on the classifier. Split the combined assertion into two hard checks per your follow-up; `!sawComplete` now fails the test on its own.

MeGaurav4:
Rebased the branch onto the v11 release (module renamed to github.com/schollz/croc/v11) so the PR is mergeable again. The hard `complete` assertion from the review is included.

schollz:
@MeGaurav4 looks like CI is failing

MeGaurav4:
Fixed the CI failure: the conflict resolution during the v11 rebase left a duplicate events import in src/cli/cli.go, src/cli/store.go and src/croc/croc.go, which broke `go build` and the Go unit tests check. Removed the duplicates (8a42e42); verified locally with `go build ./...`, `go vet` and `go test ./src/croc/ -run TestJSONEventsE2E -v` (all pass). The Go unit tests check should be green now. Could you reopen the PR? Happy to rebase again if anything else is stale.

PR #1244, opened 2026-08-15
feat: add --no-encrypt flag to skip file-content encryption

luki446:
## What

Adds a `--no-encrypt` flag that skips encrypting file **contents** while keeping the code-phrase handshake and control channel encrypted. Useful on trusted LANs or direct links where per-chunk content encryption overhead isn't wanted, without disabling the relay handshake/auth entirely.

## Behavior

- The sender's `--no-encrypt` setting drives the data path; the receiver follows it automatically — the receiver does **not** need to pass the flag.
- The encryption mode is communicated from sender to receiver via transfer metadata (`SenderInfo`).
- The receiver prints a warning that file contents travel in the clear (readable by the relay / on-path observers) before the existing Accept prompt.
- Handshake (PAKE code-phrase) and control channel remain encrypted regardless.

## Files

- `src/cli/cli.go` — new flag wiring.
- `src/croc/croc.go` — `NoEncrypt` on `Options`/`SenderInfo`, data-path conditionals, receiver warning.
- `README.md` — documentation of the flag.
- `src/croc/croc_noencrypt_test.go` — test covering both compress and no-compress branches.

## Testing

`go test ./src/croc/` — added test exercises both compression branches under `--no-encrypt`.

## Relation to #1236

Same feature, same data-path approach. Key differences here: receiver warning before accept (vs. silent decrypt-skip), cleaner `sendData` (payload built once, encrypt conditional), and README + tests included.


schollz:
what about #1236 
how does this compare to that implementation?

luki446:
Okay I wrote it few days ago and didn't notice someone made other PR implementing it. Implementation looks similar but my PR adds README mention, tests and warning for a receiving user that they are receiving unecrypted payload

schollz:
Okay cool! All good. I don't care which is used but just check in with the originator of the issue and ask if it meets their needs

luki446:
Please take a look @tropicalseal if it satisfies #1222

PR #1332, opened 2026-09-12
storecrypto: reject ChunkRef indices above uint32 range

armouldr:
## storecrypto: enforce upper bound on ChunkRef indices

`validateChunkRef` checks that `ObjectIndex`, `FileIndex`, and `FileChunk` are
non-negative, but never checks an upper bound. These fields are later
truncated to `uint32` when building the chunk's AAD in `chunkAAD`:

```go
binary.BigEndian.PutUint32(data[offset:], uint32(ref.ObjectIndex))
```

The AAD's job is to bind a chunk's ciphertext to a unique logical position, and that binding
silently breaks since two `ChunkRef`s whose `ObjectIndex` differs by exactly `2^32` produce
byte-identical AAD.

### Reproduction

This snippet (not included in the diff) demonstrates the collision:

```go
func TestChunkAADCollidesAcrossUint32Boundary(t *testing.T) { // passes because AADs are byte-identical.
	// ObjectIndex is truncated to uint32 when building the AAD, so two logically distinct indices
	// exactly 2^32 apart produce byte-identical associated data.
	id := "demo-transfer-id"
	honest := ChunkRef{ObjectIndex: 3, FileIndex: 0, FileChunk: 3, PlainSize: 4}
	colliding := ChunkRef{ObjectIndex: 3 + (1 << 32), FileIndex: 0, FileChunk: 3, PlainSize: 4}

	if !bytes.Equal(chunkAAD(id, honest), chunkAAD(id, colliding)) {
		t.Fatal("expected chunkAAD to collide for indices differing by 2^32")
	}
}
```

### Where this is reachable
- The native Go client (`src/storeclient/client.go`) always derives
  `ChunkRef`s from `ChunkRefs(manifest)`, which is bounded by the validated
  manifest's total size, so it doesn't seem reachable there.
- The store/relay service never constructs a `ChunkRef` at all.
- The WASM/browser bridge (`web/wasm/main.go`, `storeChunkRef`) builds a `ChunkRef` 
  from JS-supplied integers with no validation of its own before calling `SealChunk`/`OpenChunk`, 
  so it relies on `validateChunkRef` to enforce the invariant.

### Fix

Add the missing upper bound, matching the `uint32` range the AAD encoding assumes.

### Test added

`TestValidateChunkRefRejectsOutOfRangeIndex` confirms out-of-range indices
are now rejected by `validateChunkRef`.

### DIFF
--- a/src/storecrypto/storecrypto.go
+++ b/src/storecrypto/storecrypto.go
@@ -15,6 +15,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net/url"
 	"path"
 	"strings"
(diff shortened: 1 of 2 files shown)


schollz:
Thanks!!

schollz:
@armouldr it's failing ci

armouldr:
Yes, I think it is because of 32-bit architectures. I cast the ints being checked to int64 so it should be fine now.

